PR #6, opened 2017-10-30
Parse all in path parameters

mildbreakfast:
Fixed an issue having multiple in path parameters. Only the first encountered was being replaced.


kamilmysliwiec:
@mildbreakfast Thanks!! Merged 🙂

PR #9, opened 2017-11-22
Fix crash when encoutering modules that have not routes

nicolaslt:
This is for usage with NestJS 4.

A bit of a hack though, I would have preferred to be able to only get the Controllers modules before reaching this code but I couldn't find a way to accomplish this. Any pointers welcome if this is possible in a better manner :smile:.


kamilmysliwiec:
Hi @nicolaslt,
Thanks for your effort. I have fixed this with really similar trick 🙂

PR #10, opened 2017-11-23
Set swagger version

nicolaslt:
Set the swagger version to 2.0 as described on [the doc](https://swagger.io/docs/specification/2-0/basic-structure/)

I assume the format is in swagger 2.0 due to how the `body` parameters are reported which I think isn't valid for swagger 3.0.

### DIFF
--- a/lib/interfaces/swagger-base-config.interface.ts
+++ b/lib/interfaces/swagger-base-config.interface.ts
@@ -1,4 +1,5 @@
 export interface SwaggerBaseConfig {
+    swagger?: string,
     info?: {
         description?: string;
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
Thanks @nicolaslt ! 🙂

nicolaslt:
Hi @kamilmysliwiec,

I just only noticed: I don't think this has been included in the new [1.1.2 deployment](https://registry.npmjs.org/@nestjs/swagger/-/swagger-1.1.2.tgz)

I've had a look at the commit history and I think this might have been a mistake?

Thanks

PR #18, opened 2017-12-04
Api response description is required

nicolaslt:
ApiResponse.description is required according to the [OpenAPI 2.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-10) specification.

This is also confirmed by trying to validate the schema through [swagger-tools](https://npmjs.com/swagger-tools])'s `validate`.


kamilmysliwiec:
Thanks @nicolaslt!

PR #25, opened 2017-12-23
Fix OpenAPI Specification inconsistencies

### LINKED ISSUE #24: bug - using required in ApiModelProperty decorator
When using the decorator `ApiModelProperty` in the following way:
```ts
export class CreateCatDto {
  @ApiModelProperty({ type: String, required: true })
  @IsString()
  readonly name: String;

  @ApiModelProperty({ type: Number, required: true })
  @IsInt()
  readonly age;

  @ApiModelProperty({ type: String, required: true })
  @IsString()
  readonly breed;
}
```
(I just added the property `required` to the swagger example )

it should add the properties `name`, `age` & `breed` to the required array of the parent in the swagger doc.

## expected:
```json
"CreateCatDto": {
    "type": "object",
    "properties": {
        "name": {
            "type": "string"
        },
        "age": {
            "type": "number"
        },
        "breed": {
            "type": "string"
        }
    },
    "required": [
        "name",
        "age",
        "breed"
    ]
}
```
## actual

```json
"CreateCatDto": {
    "type": "object",
    "properties": {
        "name": {
            "type": "string",
            "required": true
        },
        "age": {
            "type": "number",
            "required": true
        },
        "breed": {
            "type": "string",
            "required": true
        }
    }
}
```

mildbreakfast:
According to OpenAPI Specification:
1. basePath must start with "/".
2. response should not contain additional properties (isArray).
3. required fields should be an array in Schema Object (resolves #24).

### DIFF
--- a/lib/document-builder.ts
+++ b/lib/document-builder.ts
@@ -44,7 +44,7 @@ export class DocumentBuilder {
   }
 
   public setBasePath(basePath: string): this {
-    this.document.basePath = /^\/.*/.test(basePath) ? basePath : '/' + basePath;
+    this.document.basePath = basePath.startsWith('/') ? basePath : '/' + basePath;
     return this;
   }
 
--- a/lib/document-builder.ts
+++ b/lib/document-builder.ts
@@ -1,8 +1,9 @@
-import { documentBase } from './fixtures/document.base';
 import {
   SwaggerBaseConfig,
   SwaggerScheme,
 } from './interfaces/swagger-base-config.interface';
+
+import { documentBase } from './fixtures/document.base';
 
 export class DocumentBuilder {
   private readonly document: SwaggerBaseConfig = documentBase;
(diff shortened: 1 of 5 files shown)


lib/document-builder.ts
```diff
@@ -43,7 +44,7 @@ export class DocumentBuilder {
   }
 
   public setBasePath(basePath: string): this {
-    this.document.basePath = basePath;
+    this.document.basePath = /^\/.*/.test(basePath) ? basePath : '/' + basePath;
```
mohsen1: use the standard `String.prototype.strartsWith`;

```ts
basePath.startsWith('/') ? basePath : '/' + basePath
 ```

kamilmysliwiec:
looks cool 🙂 will pull & test locally.
thanks @mildbreakfast ❤️

Issue #28, opened 2018-01-07
Support for @Body('property') decorator


tma-isbx:
Hi,

I've been playing around with the Swagger module. Looks great!

I did notice that it doesn't seem to support the `@Body('property')` decorator.

```
  @Post('login')
  public async login(
    @Body('email') email: string,
    @Body('password') password: string
  ): Promise<AccessToken> {
    return this.userService.login(email, password);
  }
```
Ends up generating a Swagger API without any parameters for `login`.

PR #38, opened 2018-01-23
[#28] Add support for @Body('property') decorator

tma-isbx:
This seems to work for my use cases.

Though might be a bit odd in the scenario where there are both `@Body()` and `@(Body('property')` decorators. Not sure if that even makes sense.


tma-isbx:
Is there any guidance/direction you can provide for a pull request to fulfill issue #28 "Support for @Body('property') decorator"?

PR #55, opened 2018-02-22
feat(@nestjs/swagger) add oauth2 to securityDefinitions

arkraft:
Add OAuth2 to security definitions according to [Swagger 2.0 Docs](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-definitions-object).


kamilmysliwiec:
Great job @arkraft, thanks!

PR #58, opened 2018-02-27
Add support for @Headers('property') parameter decorator.

mrgusev:
@kamilmysliwiec it is a simple change and works perfectly. Is there any reason for not including it to the map?

image

### DIFF
--- a/lib/explorers/api-parameters.explorer.ts
+++ b/lib/explorers/api-parameters.explorer.ts
@@ -136,6 +136,7 @@ const mapParamType = (key: string): string => {
         case RouteParamtypes.BODY: return 'body';
         case RouteParamtypes.PARAM: return 'path';
         case RouteParamtypes.QUERY: return 'query';
+        case RouteParamtypes.HEADERS: return 'header';
         default: return DEFAULT_PARAM_TOKEN;
     }
 };


kamilmysliwiec:
Great! thanks @mrgusev

AragurDEV:
Is there a way to disable this?
My function uses the user-agent to store it in the database.
As of now there is a user-agent field shown in swagger (which is required)

Mark-McCracken:
+1
I'd would also like to be able to disable this for a given header.
For example most parts of my API require a token in the header for authentication.
I have some middleware that parses the token, and adds some useful information to the headers, like users' groups:
```typescript
@Injectable()
export class TokenValidMiddleware implements NestMiddleware {

  constructor(private tokensService: ServerTokensService) {}

  resolve() {
    return async (req, res, next) => {
      let token = req.header(APP_TOKEN);
      if (!token) throw new UnauthorizedException(`No token provided. Log in and get a token.`);

      let details = await this.tokensService.parseToken(token);
      if (!details) throw new UnauthorizedException(`Token invalid. Login and get a new one.`);
      req.headers['user'] = details.name;
      req.headers['groups'] = this.getGroups(details);
      next();
    };
  }

  getGroups(details) {
	  if (!details.groups || !details.groups.length) return [];
  	return details.groups.map((group: string) => group.replace('some.prefix.', ''));
  }

}
```


So in my middleware, the method requires a header token, or it'll throw an unauthorised exception.
Then in my controller I have something like so:
```typescript
  @Get('group-info')
  async getGroupInfo(@Headers('user') user: string,
                     @Headers('groups') groups: string[] = []) {
    const query = `SELECT * FROM table WHERE group IN (${this.allowableGroups(groups)}) or user = ?`
    const results = await this.runQuery(query, [user]);
    return results;
  }
```

This works great and I love it, but the swagger docs shows this endpoint requiring headers that don't really make any sense to the API user:

image

The user shouldn't know about these, and has no control over them anyway.

For now, I can go through the doc, and manually edit it, but would be nice to be able to do this.

However, I think the best place to add that configuration would be in the header decorator which is in a different npm package, which might be difficult.

Do you guys have any thoughts?

AtoianAvetik:
```typescript
...
 @ApiImplicitHeader({name: 'groups', required: false})
 @ApiImplicitHeader({name: 'user', required: false})
...
```

PR #52, opened 2018-02-09
Add support for swagger customization options

fwoelffel:
**TLDR;** This should resolve https://github.com/nestjs/swagger/issues/42

---

This PR adds the support for the `swagger-ui-express` customization options defined here: https://github.com/scottie1984/swagger-ui-express/blob/master/index.js#L13

You should now, for example, be able to customize your SwaggerUI CSS:
```js
import { NestFactory } from '@nestjs/core';
import { SwaggerModule, DocumentBuilder } from '@nestjs/swagger';
import { ApplicationModule } from './app.module';

async function bootstrap() {
  const app = await NestFactory.create(ApplicationModule);

  const options = new DocumentBuilder()
    .setTitle('Cats example')
    .setDescription('The cats API description')
    .setVersion('1.0')
    .addTag('cats')
    .build();
  const document = SwaggerModule.createDocument(app, options);
  SwaggerModule.setup('/api', app, document, {
    customCss: "<your CSS rules>"
  });

  await app.listen(3001);
}
bootstrap();
```

I may add some documentation to the code in the days to come.

PR #61, opened 2018-03-06
added additional parameters to be able to use customjs for auth0 authentication

jestersimpps:


jestersimpps:
oops, quite the same: https://github.com/nestjs/swagger/pull/52

PR #53, opened 2018-02-09
Add JSON Schema Object properties support to the ApiModelProperty decorator

fwoelffel:
~~**DISCLAIMER** Looks like a duplicate of https://github.com/nestjs/swagger/pull/44~~

**TLDR;** This PR adds the `enum` property to the `@ApiModelProperty()` decorator

_This PR might also solve https://github.com/nestjs/swagger/issues/19_

---

This PR slightly updates the `@ApiModelProperty()` options by adding the `enum?: string[] | number[] | (string|number)[]`.

Here is a pretty self explanatory example:
```js
export class SampleDto {

  @ApiModelProperty({
    enum: ["String1", "String2"]
  })
  prop1: string;

  @ApiModelProperty({
    enum: ["String1", 2]
  })
  prop2: string;

  @ApiModelProperty({
    enum: [1, 2]
  })
  prop3: number;

}
```

image

PR #62, opened 2018-03-06
Add extended support for model definitions and response

sheremet:
# Add extended support for model definitions and response

- Added all supported properties for api-model-property decorator
- Added support for Api response `example` field

This changes takes ability to create completely supported Swagger definitions(used in production on our project). Will be helpful using not forked version of `@nest/swagger`


kamilmysliwiec:
This PR is outdated due to #53. Anyway, thanks for your effort!!

PR #76, opened 2018-04-30
added easier enum declaration, corrected wrong swagger for enum arrays

DavisJaunzems:
The current solution for enums doesn't generate the correct swagger specification when they are used as arrays.

```
@ApiModelProperty({
    enum: EDummyEnum,
    isArray: true
})
```
Would produce:
```
"type": "array",
"enum": [ 
    "Test1",
    "Test2"
],
"items": {`
    "type": "array"
}
```

but it should be:
```
"type": "array",
"enum": [ 
    "Test1",
    "Test2"
],
"items": {`
    "type": "string"
}
```

This Pull Request has a fix for that and it also allows to use enum as a type in the `ApiModelPorperty` rather that defining all enum values.


kamilmysliwiec:
Great @DavisJaunzems, thanks 👍

PR #74, opened 2018-04-19
fixed items.type when main type is array

testinaweb:

PR #77, opened 2018-05-03
Add support of collectionFormat property

gldmtr:
The parameter allows to change serialisation strategy, which is useful for passing arrays in a querystring. It is documented on https://swagger.io/docs/specification/2-0/describing-parameters/ under "Array and Multi-Value Parameters" header
before:
image
after I added `collectionFormat: 'multi'`:
image


malbertSC:
I think this is the same as https://github.com/nestjs/swagger/pull/74

gldmtr:
@malbertSC Not really, #74 adds support the same parameter but in another decorator

PR #91, opened 2018-05-23
Update validateRoutePath to use Express's path-to-regexp lib

crdrost:
Right now the [Express docs][route-parameters] contain paths which are not properly handled by this swagger extension, for example:

>      Route path: /plantae/:genus.:species
>      Request URL: http://localhost:3000/plantae/Prunus.persica
>      req.params: { "genus": "Prunus", "species": "persica" }
 
generates an OpenAPI doc for the path `/plantae/{genus.:species}`. The Express route still works but tools like `swagger-ui` have problems.

The underlying Express code [indicates that this responsibility was packaged into its own library][document-import], the package [path-to-regexp](https://www.npmjs.com/package/path-to-regexp), which already does all of the "remove the trailing regexp" difficult stuff for you. It seems like the proper way for this project to proceed is, rather than trying to hack up something which does what Express does, to just depend on this library as well and use *it* to break up the path string, then turn *that* into a proper Swagger path. This PR contains a first stab at that.

[route-parameters]: https://expressjs.com/en/guide/routing.html#route-parameters
[document-import]: https://github.com/expressjs/express/blob/master/lib/router/layer.js#L16

### DIFF
--- a/lib/swagger-explorer.ts
+++ b/lib/swagger-explorer.ts
@@ -28,6 +28,7 @@ import { MetadataScanner } from '@nestjs/core/metadata-scanner';
 import { RequestMethod } from '@nestjs/common';
 import { exploreApiOperationMetadata } from './explorers/api-operation.explorer';
 import { exploreApiParametersMetadata } from './explorers/api-parameters.explorer';
+import * as pathToRegexp from 'path-to-regexp';
 
 export class SwaggerExplorer {
   private readonly metadataScanner = new MetadataScanner();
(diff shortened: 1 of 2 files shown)


crdrost:
I don't see a proper test suite so I wanted to share what this does a little more directly. [Here is a gist for a side-by-side runner](https://gist.github.com/crdrost/347bc0613860c0c67644ccb4bcfdb748). It produces:

```
# Old Style
  /hi/:hows/:it/:goin                =>   /hi/{hows}/{it}/{goin}
  sorry/you/couldnt/:get.:through    =>   /sorry/you/couldnt/{get.:through}
  please/leave/your/name             =>   /please/leave/your/name
  /                                  =>
  and/                               =>   /and/
  /your/:number(d+)                  =>   /your/{number}
                                     =>   /
  /and/we'll/:get.:back/:to/:you     =>   /and/we'll/{get.:back}/{to}/{you}

# New Style
  /hi/:hows/:it/:goin                =>   /hi/{hows}/{it}/{goin}
  sorry/you/couldnt/:get.:through    =>   /sorry/you/couldnt/{get}.{through}
  please/leave/your/name             =>   /please/leave/your/name
  /                                  =>
  and/                               =>   /and/
  /your/:number(d+)                  =>   /your/{number}
                                     =>   /
  /and/we'll/:get.:back/:to/:you     =>   /and/we'll/{get}.{back}/{to}/{you}
```

crdrost:
@kamilmysliwiec what else would I need to do to get this bugfix committed and up on NPM? My team would really like this and I'd rather not switch to an `@crdrost/nest-swagger` type of NPM import if I don't have to.

kamilmysliwiec:
Well done! Thanks 🙂

Issue #92, opened 2018-05-24
Feature Request: Turn off swagger for methods/controllers


nursik:
I did not find how to turn off swagger for controllers/methods. It could be a good feature. If there is any way to turn off swagger, I will be grateful for explaining ho to do that.

kamilmysliwiec:
PR has been merged. This feature will be available in the next release (v2.1.0)

yanarp:
Hi, I didn't find a way to hide the controller in the nestjs/swagger documentation can somebody nudge me in right direction.

yanarp:
> Hi, I didn't find a way to hide the controller in the nestjs/swagger documentation can somebody nudge me in right direction.

No issues, got it ! : P

geersch:
For people with the same question, you can apply `@ApiExcludeEndpoint()` to a controller method to exclude it from the Swagger documentation.

dtstsUI:
@ ApiExcludeEndpoint() decorator only hide the endpoint for Swagger - not from an outer access.

is there a way to also hide the endpoint(by condition) to an external source?

thanks.

vanakema:
> @ ApiExcludeEndpoint() decorator only hides the endpoint from Swagger - not from an outer access.
> 
> is there a way to also hide the endpoint(by condition) from an external source?
> 
> thanks.

The swagger annotations are only for generating an API spec. There's no real way to "hide" an endpoint through your server. Technically if you didn't make a swagger API spec, all the endpoints would be unlisted, but still accessible from the outside world. 

To solve your problem, you need to figure out a way to programmatically differentiate between requests from external and internal sources (ex, using the AuthGuard, where you assume that all authenticated requests are internal, and all external requests, you return a 404). You could also somehow firewall off those endpoints to IP ranges that are not part of your internal network.

dtsanan:
> > @ ApiExcludeEndpoint() decorator only hides the endpoint from Swagger - not from an outer access.
> > is there a way to also hide the endpoint(by condition) from an external source?
> > thanks.
> 
> The swagger annotations are only for generating an API spec. There's no real way to "hide" an endpoint through your server. Technically if you didn't make a swagger API spec, all the endpoints would be unlisted, but still accessible from the outside world.
> 
[...]

PR #93, opened 2018-05-24
Disable endpoint option

Ob-La-Di:
Hi,

I made another decorator to disable an endpoint, to fit my company needs. It just doesn't add the endpoint to swagger-ui for the routes with the decorator.

To use it, you juste need to add `ApiDisableEndpoint()` before a controller method.

I'm still a junior dev, so if there's any mistakes, feel free to let me know and I'll gladly correct it.

Bonus: It coincidentally resolves issue #92.

Thanks


Ob-La-Di:
Changed the name of the decorator and made a PR to the documentation. 👍

thaoula:
Hi @kamilmysliwiec,

Do you think this PR will be merged? 

Currently we are doing a bit of gymnastics with multiple nest applications to try and document only our public API using the swagger module.

Would love to just use an attribute :)

nartc:
Though it might be a hassle but `@ApiHidden()` would be a great name also.

kamilmysliwiec:
Amazing job, thanks @Ob-La-Di !

PR #163, opened 2018-11-22
feat: set operationId to method name by default

arkraft:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

No operationId is set, if you want to use it, you have to manually define one by using the `ApiOperation` decorator. 

## What is the new behavior?
By default, the operationId is the method name. You can still use the `ApiOperation` decorator which will overwrite the default value.

```
[x] Yes
[ ] No
```

## Other information

### DIFF
--- a/lib/swagger-explorer.ts
+++ b/lib/swagger-explorer.ts
@@ -158,7 +158,8 @@ export class SwaggerExplorer {
     const fullPath = globalPath + this.validateRoutePath(routePath);
     return {
       method: RequestMethod[requestMethod].toLowerCase(),
-      path: fullPath === '' ? '/' : fullPath
+      path: fullPath === '' ? '/' : fullPath,
+      operationId: method.name
     };
   }
 


villelahdenvuo:
It's definitely a breaking change because most client library generators use the operation id as the method name in the generated library.

I created my own decorator that simply sets the default operationId and calls the nest decorator, pretty easy. But I do think it should do this by default because using the path leads into weird/unusable ids.

nartc:
```ts
export const ApiOperationId = (options?: {
  title?: string;
  description?: string;
  operationId?: string;
  deprecated?: boolean;
}) => {
  return (target: any, propertyKey: string, descriptor: PropertyDescriptor) => {
    const controllerName = target.constructor.name;

    const metadata = options
      ? { ...options, title: options.title || propertyKey }
      : { title: propertyKey };

    ApiOperation({
      ...metadata,
      title: metadata.title,
      operationId: `${controllerName.substr(
        0,
        controllerName.indexOf('Controller'),
      )}_${metadata.title}`,
    })(target, propertyKey, descriptor);
  };
};

```

This is my own custom decorator that will extract the controller method name to:
`{Class}_{methodName}` for the `operationId`. Additionally, you can always override the `methodName` by passing in `title` for the decorator's options.

```ts
@ApiOperationId()
  async updateAvatar(
    @CurrentUser() currentUser: AuthUser,
    @UploadedFile() file: CloudinaryFile,
  ): Promise<UserVm> {
    return this.userService.updateAvatar(currentUser, file);
  }
```
This will give me: `User_updateAvatar` for `operationId`. 

PS: I am using NSwag to generate TS clients and the convention is: `{Class}_{method}`

arkraft:
done in #355

Issue #372, opened 2019-11-03
Reusable enums support


flogh:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Currently enums are not defined under components section

```
"components": {
    "schemas": {
        "Album": {
            "type": "object",
            "properties": {
                "direction": {
                    "type": "string",
                    "enum": [
                        "Up",
                        "Down",
                        "Right",
                        "Left"
                    ]
                }
            }
        }
    }
}
```

## Expected behavior

```
"components": {
    "schemas": {
        "direction": {
            "type": "string",
            "enum": [
                "Up",
                "Down",
                "Right",
                "Left"
            ]
        },
        "Album": {
            "type": "object",
            "properties": {
                "direction": {
                    "schema": {
                        "$ref": "#/components/schemas/direction"
                    }
                }
            }
        }
    }
}
```

## What is the motivation / use case for changing the behavior?

OpenApi 3 allows us to reference enums under components section to use them multiple times. 
See the "Reusable Enums" section here [https://swagger.io/docs/specification/data-models/enums/](https://swagger.io/docs/specification/data-models/enums/)

It would allow us to get enums types on the front side too with classes, using a generator like ng-openapi-gen for example, which would be amazing.

## Environment

<pre><code>
Nest version: 4.0.0-next.2

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

nartc:
There is one caveat about this feature which is getting the Enum name at runtime. One possible solution without adding more property to `ApiProperty` is to have: `enum: ` to accept a `lazily-evaluated function` as follow:
```ts
enum Status {
   Inactive = 'inactive',
   Active = 'active'
}

@ApiProperty({enum: () => Status})
status: Status;
```

This will allow us to grab the `Status` string as the enum name at runtime to be able to "reuse" it. Any suggestions?

flogh:
It sounds good to me

nartc:
PR #412 is in

nartc:
@flogh are you open to pull my branch and give it some test runs?

flogh:
Of course ! I'm doing it right now

nartc:
[...]

PR #412, opened 2019-12-04
feature(): extract Enum out to its schema

nartc:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Current behavior is when an `Enum` is defined and used on a `Model`, the `Enum` values are put on the `properties` of the `Model` itself.
screen shot 2018-06-28 at 7 40 45 pm

This causes Frontend Generation Tools (such as *NSwag*) to generate multiple `enums` even though they are meant for the same thing.
screen shot 2018-06-28 at 7 38 12 pm

Issue Number: #372 

## What is the new behavior?
~~New behavior will expect: `enum` property on various decorator (`ApiParam`, `ApiQuery`, `ApiProperty`) to accept a lazy-function syntax.~~

~~This allows `SwaggerExplorer` to grab the Enum name at run-time which makes it possible to extract the Enum out to `schemas` to be reused on `Parameters`~~

~~Please note this is an **addition**, not **replacement**. `enum` property still accepts what it currently accepts. This PR makes it so it will solve the issue stated in `Current Behavior` with `lazy function` syntax.~~

Various decorators that accept `enum` now also accept `enumName`. If you want your `enum` to be store as a schema (reusable) then you need to provide `enumName` as well. This implementation might seem a little bit explicit compared to the `lazy load function` but its intention is also clearer. 

```typescript
// before
@ApiProperty({enum: Role})

// later
@ApiProperty({enum: Role, enumName: 'Role'})
```

Again, although the PR is working, maintain the current passing tests, passing  two new tests, my implementation might contain flaws and might go against NestJS's design mental. That said, I'm open to ALL suggestions and feedbacks.

```
[ ] Yes
[x] No
```

## Other information


flogh:
Ok , so I’ve just finished updated my project, which is quite big, I had about 60 enums. Some where in entities, in responses types, in queries, params. 

Every single one has correctly been defined under the components section, and reused where needed with the « $ref ».

But I've noticed something undesirable :

- If the enum is defined in the same file as the class :

```
export enum Direction {
    UP = "Up",
    DOWN = "Down",
    LEFT = "Left",
    RIGHT = "Right",
}

export class Album {
    @ApiProperty({ enum: () => Direction })
    public direction: Direction;
}
```
It behaves normally :
```
"components": {
    "schemas": {
        "Direction": {
            "type": "string",
            "enum": [
               "Up",
               "Down",
               "Left",
               "Right"
           ]
        },
         "Album": {
             "type": "object",
             "properties": {
                 "direction": {
                     "$ref": "#/components/schemas/ValidTypes"
                 }
             }
        }
    }
}
```

- If the enum is imported in the class file :

```
import { Direction } from "whatever/my-enums";

export class Album {
    @ApiProperty({ enum: () => Direction })
    public direction: Direction;
}
```
It behaves like this :
```
"components": {
    "schemas": {
        "my_enums_1.Direction": {
            "type": "string",
            "enum": [
               "Up",
               "Down",
               "Left",
               "Right"
           ]
        },
         "Album": {
             "type": "object",
             "properties": {
                 "direction": {
                     "$ref": "#/components/schemas/my_enums_1.Direction"
                 }
             }
        }
    }
}
```

The enum name is prefixed by the file's name where it's imported

Would you be able to solve this issue ?

nartc:
@flogh Thanks for testing. It's super helpful. The unexpected behavior is kind of expected but I wasn't sure. I am able to solve the issue but it's flaky. I'll update the PR but keep in mind that @kamilmysliwiec might have other suggestions. We might actually need to introduce a new `property` for some decorators instead of rely on `lazy function` syntax.

flogh:
Ok, no problem :)

By the way, nice job ! 👍

nartc:
@flogh `getEnumName` function has been modified. Please try again if you can.

flogh:
Ok, working as expected 👏

kamilmysliwiec:
Perhaps I missed something, but how can I control whether `enum` should be defined as a component or defined "inline" instead?

flogh:
As @nartc said, this PR makes enums behave like this : 

- Enum defined inline 
```
@ApiProperty({ enum: Role })
```

- Enum defined under components 
```
@ApiProperty({ enum: () => Role })
```

He supposed that we might introduce a new decorator like so :
- Enum defined under components 
```
@ApiProperty({ enum: Role, asComponent: true })
```
But I don't know if it's possible

lib/document-builder.ts
```diff
@@ -60,14 +60,16 @@ export class DocumentBuilder {
...
+    this.document.tags = this.document.tags.concat(
+      pickBy(
+        {
+          name,
+          description,
+          externalDocs
+        },
+        negate(isUndefined)
+      ) as TagObject
+    );
```
kamilmysliwiec: Can we create a separate PR for this? :)

nartc: Oh that’s a “npm run format” but yes, of course we can.

nartc:
> Perhaps I missed something, but how can I control whether `enum` should be defined as a component or defined "inline" instead?

@kamilmysliwiec My intention for this PR is to **not** require users to change anything unless they want to have `Enum` defined as a `Schema` to be reusable (per OpenAPI Spec). Currently, we have:

```typescript
// enum
@ApiProperty({ enum: SomeEnum })

// inline
@ApiProperty({ enum: ['a', 'b', 'c'] })
```

These two will stay working as expected, no `Schema` will be created for `Enum`. 

This PR added a lazy-function syntax to `enum` as follow:

```typescript
@ApiProperty({ enum: () => SomeEnum })
```

The lazy-function allows `SwaggerExplorer` to grab the `Enum` name at runtime. If the users pass in a lazy-function for their `enum` then I handle and construct an internal `enumName` on the `param` then `delete` that before it gets to being added to `OpenAPI Spec` file. 

I am aware that `() => SomeEnum` is flaky but in order to prevent adding an additional field to these existing decorators: `ApiProperty`, `ApiParam`, `ApiQuery` (maybe `ApiBody`, I might have missed something), I went with `() => SomeEnum` instead of `ApiProperty({ enum: SomeEnum, enumName: 'SomeEnum' })`

lib/document-builder.ts
nartc: @kamilmysliwiec do you still want this formatting in a separate PR?

nartc: removed formatting change

nartc:
unit tests and e2e test have been updated.

flogh:
I'm using this PR for more than a month now on my project, and it's working great. Would it be able to merge it on the master branch ?

kamilmysliwiec:
@nartc can we somehow add a way to more explicitly control whether I want to extract an enum or not (additional property)? Instead of implicitly assuming that every lazy-function = extract an enum.

nartc:
@kamilmysliwiec Implicitly based on the `lazy-function` is to just allowing TS to extract the Enum name at run time. I didn't want to add more options to `ApiProperty`, that's why I just used `lazy-function`. If we are able to make it clear that using `lazy-function` to declare an `enum` on `ApiProperty`, then it will be extracted. 

With that said, if it's still essential that we need to explicitly control the extraction, then instead of relying on `lazy-function`, I can just add another property called `enumName` to `ApiProperty`. 

You have any suggestions?

kamilmysliwiec:
> With that said, if it's still essential that we need to explicitly control the extraction, then instead of relying on lazy-function, I can just add another property called enumName to ApiProperty.

Even though this may sound like an additional, extra overhead, I think this will be more straightforward for people.

nartc:
@kamilmysliwiec Alrighty then. It makes the PR even simpler with `enumName`. 
@flogh Since it's likely that the API is going to change for the Enum extraction, there will be changes to your code if the new change (what Kamil and I are talking about) lands on master.

flogh:
@nartc No prob ! 👍

prevostc:
Would be a great addition!

nartc:
`enum` is no longer taking in a lazy function to "guess" the `enumName`. Instead, `enumName` is a separate property to be passed in if the consumers want to have this `enum` to be extracted out to its separate schema. 

```typescript
// beginning of this PR
@ApiProperty({ enum: () => Role }) // Role will be extracted out

// now
@ApiProperty({ enum: Role, enumName: 'Role' }) // Role will be extracted out
```

The motivation for this change was in the discussion in this PR. Mainly, we don't want to change the behavior of existing API, so introducing `enumName` as an addition property to the `options` is a better trade-off.

kamilmysliwiec:
@nartc  would you like to create a PR to the docs? :)

nartc:
@kamilmysliwiec Will do in couple of hours.

kamilmysliwiec:
@nartc awesome, thank you!

kamilmysliwiec:
Thanks!

PR #452, opened 2019-12-26
fix(): ApiProperty function type not work

fourcels:
## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Query Parameters function type not work
```typescript
export class QueryParam {
  @ApiProperty({type: () => Number})
  page: number;
}
```
image

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/lib/services/swagger-types-mapper.ts
+++ b/lib/services/swagger-types-mapper.ts
@@ -13,7 +13,9 @@ export class SwaggerTypesMapper {
       const { type } = param as ParamWithTypeMetadata;
       const typeName =
         type && isFunction(type)
-          ? this.mapTypeToOpenAPIType(type().name)
+          ? this.mapTypeToOpenAPIType(
+              type.name === 'type' ? type().name : type.name
+            )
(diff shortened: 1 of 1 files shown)


lib/services/swagger-types-mapper.ts
```diff
@@ -13,7 +13,7 @@ export class SwaggerTypesMapper {
       const { type } = param as ParamWithTypeMetadata;
       const typeName =
         type && isFunction(type)
-          ? this.mapTypeToOpenAPIType(type.name)
+          ? this.mapTypeToOpenAPIType(type().name)
```
kamilmysliwiec:
```
type && isFunction(type)
```
will give `true` for type references (class = function) too. We can't simply evaluate this

kamilmysliwiec:
Thank you for your contribution!

Tests are failing now. Also, please, take a look at my comment above

kamilmysliwiec:
This has been fixed in 4.3.2 release

Issue #482, opened 2020-01-13
Default Operation ID clashes 


wafs:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Two controllers with the same operation will produce the same operation ID if no ID is provided when using tsc to compile. 

```typescript
@Controller('a')
class A {
    @ApiOperation()
    getOne(){...}
}

@Controller('b')
class B {
    @ApiOperation()
    getOne(){...}
}
```

```json
// Output truncated 
{
    "paths" : {
        "a" : {"get" : {"operationId": "getOne"}},
        "b" : {"get" : {"operationId": "getOne"}},
    }
}

```

## Expected behavior

Take the name of the controller also, or append a hash 

```json
// Output truncated 
{
    "paths" : {
        "a" : {"get" : {"operationId": "a__getOne"}},
        "b" : {"get" : {"operationId": "b__getOne"}},
        // or         
        "a" : {"get" : {"operationId": "getOne_r@nd0mH4$h"}},
        "b" : {"get" : {"operationId": "getOne_abcd12345"}},

    }
}

```

## Minimal reproduction of the problem with instructions

1. Create two controllers with the same name for a method
2. Expose the same named method with `@ApiOperation()` decorator, do not provide an `operationId`
3. Use TSC to compile (untested if `nest build` command works)
4. View the JSON output 

## What is the motivation / use case for changing the behavior?
This is causing issues where operationId fields are expected to be globally unique but the value of them isn't important. I am using the JSON output in an OpenAPI UI generator called [redoc](https://github.com/Redocly/redoc). Without the unique operationIds, the links it produces are broken because all links with the same name point to the same endpoint, and this is frequent for my project as I'm using generic controllers. 

## Environment

Nest version: 6.10.14
 
For Tooling issues:
- Node version: 11.15
- Platform:  Mac

wafs:
Currently using this as a bandaid fix after the json doc compiles:

```typescript
Object.entries(document.paths).forEach(([path, obj]) => {
    ['get', 'patch', 'put', 'post', 'delete'].forEach(method => {
      if (obj[method]) {
        obj[method].operationId = `${path.split('/')[1]}__${obj[method].operationId}`;
      }
    });
  });
```

sangamk:
[...]

PR #487, opened 2020-01-14
fix(): use unique operation id for each endpoint (class_method)

kamilmysliwiec:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Operation ID == method name

Issue Number: #482 

## What is the new behavior?

Operation ID == ClassName_MethodName

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/lib/swagger-explorer.ts
+++ b/lib/swagger-explorer.ts
@@ -205,8 +205,15 @@ export class SwaggerExplorer {
     return {
       method: RequestMethod[requestMethod].toLowerCase(),
       path: fullPath === '' ? '/' : fullPath,
-      operationId: method.name
+      operationId: this.getOperationId(instance, method)
     };
+  }
+
+  private getOperationId(instance: object, method: Function): string {
+    if (instance.constructor) {
+      return `${instance.constructor.name}_${method.name}`;
+    }
+    return method.name;
   }
 
   private reflectControllerPath(metatype: Type<unknown>): string {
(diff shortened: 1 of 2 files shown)


lib/swagger-explorer.ts
```diff
@@ -205,10 +205,17 @@ export class SwaggerExplorer {
...
       method: RequestMethod[requestMethod].toLowerCase(),
       path: fullPath === '' ? '/' : fullPath,
-      operationId: method.name
+      operationId: this.getOperationId(instance, method)
     };
   }
 
+  private getOperationId(instance: object, method: Function): string {
+    if (instance.constructor) {
+      return `${instance.constructor.name}_${method.name}`;
```
boenrobot: As much as I need this fix too, wouldn't it be better if an invalid JS character is used as a separator, to prevent creating unintentional clashes?

Like "-" for example? Generators should be smart enough to escape that into the target language, and doc generators can generate URLs without requiring URL escaping even.

Example of a clash:

ModuleController_UsingThisCase controller
DoAction method

vs.
ModuleController controller
UsingThisCase_DoAction method

sure, chances of such collisions happening in the same code base are minimal, at best, given the popularity of linting tools, but still.

kamilmysliwiec: I'm not sure if we should protect ourselves against code that shouldn't appear in any codebase since it simply violates too many rules (+ probability of having such a naming clash can be brought down to 0)

Syarx:
This is a breaking change. You should give an option to disable it. Some of us have to rewrite hundreds of calls now. I have to downgrade until then sadly

PR #728, opened 2020-05-08
fix(): process isarray for string based property type

BasKiers:
## PR Checklist

- [ x ] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ x ] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ x ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Using the following options: 
```
@ApiProperty({ type: 'integer', isArray: true })
someProp: number[];
```
Produces the following swagger parameter definition, ignoring the isArray flag:
```
"someProp": {
  "type": "integer",
},
```

Issue Number: N/A

## What is the new behavior?
Now, when the type option is a string, the isArray property will still be processed, fixing the above issue:
```
@ApiProperty({ type: 'integer', isArray: true })
someProp: number[];
```
Produces the following swagger parameter definition:
```
"someProp": {
  "type": "array",
  "items": {
    "type": "integer"
  }
},
```

```
[ ] Yes
[ x ] No
```

## Other information

### DIFF
--- a/lib/services/schema-object-factory.ts
+++ b/lib/services/schema-object-factory.ts
@@ -43,7 +43,7 @@ export class SchemaObjectFactory {
     schemas: SchemaObject[],
     schemaRefsStack: string[] = []
   ): Array<ParamWithTypeMetadata | BaseParameterObject> {
-    return parameters.map(param => {
+    return parameters.map((param) => {
       if (!isBodyParameter(param)) {
         return this.createQueryOrParamSchema(param, schemas, schemaRefsStack);
       }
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
Thanks for your contribution!

TBH I'm not certainly sure what to do about this PR. If you pass a type as a string, we don't modify it on purpose. However, your improvement might be useful when specifying types like `integer` (as in your example) which are impossible to represent using type refs (`type: Number`)

BasKiers:
The only change this PR adds is that, when passing a string as the `type` property, the `isArray` property is no longer ignored. The current behavior, where `isArray` is handled in all cases except for when the type is a string was confusing and prevents us from generating the desired swagger with `type: 'integer'` and `isArray: true` since there is no other way to give a raw definition to the parameter property.

Issue #1112, opened 2020-12-27
Arrays and objects are not managed properly when located in first level of @Query()


arnaud-cortisse:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior
I created these DTOs:
```
class ObjectDto {
  @ApiProperty()
  field: string;
}

export class HelloWorldDto {
  @ApiProperty({
    isArray: true,
    type: ObjectDto,
  })
  arrayOfObjectsDto: ObjectDto[];
}
```
And expected them as query parameters:
```
  @Get('/Not Working')
  getHello(@Query() helloWorldDto: HelloWorldDto): string {
    return this.appService.getHello();
  }

```
I got the following output:
![image](https://user-images.githubusercontent.com/28689103/103168510-0daef080-4834-11eb-93d0-de47f48bc577.png)

## Expected behavior

I would expect something like this:
![image](https://user-images.githubusercontent.com/28689103/103168519-2cad8280-4834-11eb-926e-4fe368882b4f.png)
As you can see, it's an actual array of objects. 

I was able to get it to work by doing the following:

```
class ObjectDto {
  @ApiProperty()
  field: string;
}

export class Params {
  @ApiProperty({
    isArray: true,
    type: ObjectDto,
  })
  arrayOfObjectsDto: ObjectDto[];
}

export class HelloWorldDto2 {
  @ApiProperty()
  params: Params;
}
```
As you can see, arrays seem to be declared correctly when they are located in the deeper level in the DTO. 
However, this is not a viable option, since the path HelloWorldDto2.params actually doesn't exist...

## Minimal reproduction of the problem with instructions

See https://github.com/arnaud-cortisse/nestjs-openapi-array-of-objects-bug

## What is the motivation / use case for changing the behavior?

I want to be able to declare arrays of complex objects in the first level of DTOs

## Environment

<pre><code>
Nest version: 7.5.1

 
For Tooling issues:
- Node version: node:alpine docker image
- Platform: node:alpine docker image

kamilmysliwiec:
Would you like to create a PR for this issue?

arnaud-cortisse:
@kamilmysliwiec I've just created a PR here: https://github.com/nestjs/swagger/pull/1113 :+1:

kamilmysliwiec:
[...]

PR #1113, opened 2020-12-28
fix: Arrays and objects are not managed properly when located in first level of @Query()

arnaud-cortisse:
## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

See https://github.com/nestjs/swagger/issues/1112

## What is the new behavior?

Arrays and objects are now correctly interpreted as such in @Query()

image

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/lib/services/schema-object-factory.ts
+++ b/lib/services/schema-object-factory.ts
@@ -62,36 +62,40 @@ export class SchemaObjectFactory {
         return this.createQueryOrParamSchema(param, schemas, schemaRefsStack);
       }
 
-      const modelName = this.exploreModelSchema(
-        param.type,
-        schemas,
-        schemaRefsStack
-      );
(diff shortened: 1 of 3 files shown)


andrewdhazlett:
@arnaud-cortisse does this work properly with the `Try it out` feature of Swagger UI? It looks like the [Swagger docs](https://swagger.io/docs/specification/describing-parameters/#schema-vs-content) want you to use a `content` block along with encoding types to specify JSON query strings. 

Here's a workaround I've been using in the meantime, but unfortunately it doesn't let me associate it to the DTO (I'd like to avoid using `$ref` here if possible)  
image

abdulkaderelrawas:
any news about this PR?

Issue #1130, opened 2021-01-12
ApiExtension incorrectly handles Array values


HMilbradt:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Using the `ApiExtension` decorator the way the code is currently written prevents you from using an array as the value of the extension.

Examples:

```
  @ApiExtension('x-codeSamples', [
    {
      lang: 'Bash',
      source: 'echo "Hello"',
    },
  ])
```

This produces the following:

```
"x-codeSamples":{"0":{"lang":"Bash","source":"echo \"Hello\""}}
```

## Expected behavior

Array formats should be allowed, as mentioned in the Open API spec: https://swagger.io/docs/specification/openapi-extensions/

> The extension value can be a primitive, an array, an object or null.

The correct output from the above example should be:

```
"x-codeSamples":[{"lang":"Bash","source":"echo \"Hello\""}]
```

## What is the motivation / use case for changing the behavior?

Plenty of extensions use array formats for specifying input data.  One example is Redocly, which allows for the `x-codeSamples` to be provided as an array.

## Environment

<pre><code>
Nest version: 7.0.0

</code></pre>

kamilmysliwiec:
Would you like to create a PR for this issue?

HMilbradt:
Yes, I'll get something working today.  I've already taken a look at the source and found the issue, looks to be a small change

kamilmysliwiec:
Thanks! Let's track this here https://github.com/nestjs/swagger/pull/1137

PR #1137, opened 2021-01-15
fix(api-extension): support for additional api extension values

HMilbradt:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1130 

## What is the new behavior?

The `ApiExtension` decorator now accepts any value to be assigned.  As referenced [here](https://swagger.io/docs/specification/openapi-extensions/), "The extension value can be a primitive, an array, an object or null.".

```
[ ] Yes
[x] No
```

## Other information

I did a quick poke around and didn't see anywhere that a test case for this may be added and checked.  However, if there is somewhere that I should be testing this behaviour, please let me know 🙂

### DIFF
--- a/lib/decorators/api-extension.decorator.ts
+++ b/lib/decorators/api-extension.decorator.ts
@@ -9,10 +9,7 @@ export function ApiExtension(extensionKey: string, extensionProperties: any) {
   }
 
   const extensionObject = {
-    [extensionKey]:
-      typeof extensionProperties !== 'string'
-        ? { ...extensionProperties }
-        : extensionProperties
+    [extensionKey]: extensionProperties
   };
 
   return createMixedDecorator(DECORATORS.API_EXTENSION, extensionObject);


lib/decorators/api-extension.decorator.ts
```diff
@@ -9,10 +9,7 @@ export function ApiExtension(extensionKey: string, extensionProperties: any) {
   }
 
   const extensionObject = {
-    [extensionKey]:
-      typeof extensionProperties !== 'string'
-        ? { ...extensionProperties }
-        : extensionProperties
+    [extensionKey]: extensionProperties
```
kamilmysliwiec, line 12: This condition was there for a reason (I'm not sure but I believe the reason was that we wanted to make sure the original `extensionProperties` object won't get mutated and so we clone it). Instead of removing it, I'd suggest adding another "branch" of this condition, making sure we don't spread the object when it's an array (but run sth like `[].concat(extenstionProperties)` or `[...extenstionProperties]`)

jiqiang90:
Hi I'm facing the same problem, any update on this? Thanks

lib/decorators/api-extension.decorator.ts
coyotte508, line 12:
```suggestion
    [extensionKey]: clone(extensionProperties)
```

Where `clone` is imported from `lodash`.

coyotte508: This can work:

```suggestion
    [extensionKey]: clone(extensionProperties)
```

(with `clone` imported from `lodash`)

kamilmysliwiec:
Let's track this here https://github.com/nestjs/swagger/pull/1174

PR #1169, opened 2021-02-02
perf: store schemas in an object instead of an array of key/value

coyotte508:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: perf improvement
```

Currently the schemas are passed around with two variables:
- `schemas` of type `Array<{[key: string]: SchemaObject>`
- `schemasRefPaths` which lists the keys

There are a few issues:
- `schemas` is wrongly typed as `SchemaObject[]` in the code
- Two related arguments need to be passed to many function accepting schemas
- To check if an item is already in `schemas`, `schemasRefPaths.includes(key)` is used which is O(schemas.length)

## What is the new behavior?

This PR merges `schemaRefsPaths` and `schemas` into only one variable, `schemas`, and allows for fast lookup in O(1). 

A variable `pendingSchemaRefs` is added only where needed -- where the use of `schemasRefPaths` was preventing infinite recursion. It's passed by copy when modified, to stay as small as possible, and only contains the refs to schemas _in the process of_ being added to `schemas`.

The typing of `schemas` is also now accurate.
```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/explorers/api-parameters.explorer.ts
+++ b/lib/explorers/api-parameters.explorer.ts
@@ -23,8 +23,7 @@ const schemaObjectFactory = new SchemaObjectFactory(
 );
 
 export const exploreApiParametersMetadata = (
-  schemas: SchemaObject[],
-  schemaRefsStack: [],
+  schemas: Record<string, SchemaObject>,
   instance: object,
   prototype: Type<unknown>,
   method: Function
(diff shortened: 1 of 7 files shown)


kamilmysliwiec:
LGTM. Merged to 8.0.0 branch

Issue #1178, opened 2021-02-05
Generated OpenAPI declares format in the wrong place for arrays


jessemyers:
<pre><code>
[x] Regression
[ ] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

This issue was previously reported and fixed as https://github.com/nestjs/swagger/issues/172

## Current behavior

When declaring a DTO with an array-valued field, the OpenAPI format parameter gets applied to the parent object rather than to the nested `items` object:

```js
{
   "format": "date-time",
   "items": {
      "type": "string"
    },
   "type": "array"
}
```

## Expected behavior

The format should be applied within the `items` object:

```js
{
   "items": {
      "format": "date-time",
      "type": "string"
    },
   "type": "array"
}
```

## Minimal reproduction of the problem with instructions

I've created an `index.js` and `package.json` [example as a gist](https://gist.github.com/jessemyers/5b4edda9db5a1d7832c6a71c8718b9d1)

My steps to reproduce:

```sh
yarn install
node index.js
# point browser to http://localhost:3000/openapi-json
```
## What is the motivation / use case for changing the behavior?

I want to generate the correct OpenAPI spec for arrays of values.

## Environment

<pre><code>
Nest version: 7.6.11
 
For Tooling issues:
- Node version: v14.15.1
- Platform:  Mac
</code></pre>

kamilmysliwiec:
Would you like to create a PR for this issue?

jessemyers:
I'm willing to try. It'd help a lot if you can point me to a test case or set of test cases that cover this code or something close to it.

jessemyers:
@kamilmysliwiec I've created a PR. I'm new to this project's style and tooling, but hopefully I've met your standards.

kamilmysliwiec:
Thanks! Let's track this here https://github.com/nestjs/swagger/pull/1182

PR #1182, opened 2021-02-06
Array format placement

jessemyers:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1178

## What is the new behavior?

Fixes the bug.

```
[ ] Yes
[x] No
```

## Other information

The set of possible fields that should be moved from the outer type to the inner type in this case could be expanded. I picked a set of fields that match my own use cases, but these might not be exhaustive!

### DIFF
--- a/lib/services/schema-object-factory.ts
+++ b/lib/services/schema-object-factory.ts
@@ -8,7 +8,8 @@ import {
   keyBy,
   mapValues,
   omit,
-  omitBy
+  omitBy,
+  pick
 } from 'lodash';
 import { DECORATORS } from '../constants';
 import { getTypeIsArrayTuple } from '../decorators/helpers';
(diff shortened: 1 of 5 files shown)


lib/services/schema-object-factory.ts
```diff
@@ -424,15 +425,20 @@ export class SchemaObjectFactory {
     type: string | Record<string, any>
   ): SchemaObjectMetadata {
     const keysToRemove = ['type', 'enum'];
+    const keysToMove = [
+      'format',
```
jessemyers, line 429: The original bug report was only for `format`, but these other keys are all similarly applicable to the inner type and are fields that my own projects use. There may be other fields that belong here.

test/services/fixtures/create-user.dto.ts
```diff
@@ -46,7 +46,8 @@ export class CreateUserDto {
 
   @ApiProperty({
     type: String,
-    isArray: true
+    isArray: true,
+    format: 'uri'
```
jessemyers, line 50: This fixture did not previously define a `format`, but it looked like an appropriate place to demonstrate the expected behavior.

adrian-hippo (approved):

jessemyers:
Hey @kamilmysliwiec, are you expecting anything else here?

kamilmysliwiec:
LGTM

Issue #1097, opened 2020-12-17
Enhance introspectComments to provide summary and description for routes


tom-thorne:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior
The nest-cli swagger plugin with `introspectComments` set to true works very nicely adding descriptions to routes from comments.

However, it is not yet possible to add a summary or a title to a route via a comment.  Instead you have to revert to adding a decorator again which is a shame, as below.

```
    /**
     * Gets operations. Use blah blah blah to filter the operation types in the response.
     */
    @Get()
    @ApiOperation({ summary: 'Get All Operations' })
    async getOperations( @Context() context: INestContext, @Query() params: GetOperationsQuery ): Promise<GetOperationsResponse> {
```

## Desired behavior

What would be nice if the comment's 1st line could represent the summary/title, and subsequent lines the description, as below.  This removes the need for the `@ApiOperation` decorator.

```
    /**
     * Get an operation by Id
     * Gets operations. Use blah blah blah to filter the operation types in the response.
     */
    @Get()
    async getOperations( @Context() context: INestContext, @Query() params: GetOperationsQuery ): Promise<GetOperationsResponse> {
```

## Environment

Nest version: 7.5.1
@nestjs/swagger: 4.7.6

kamilmysliwiec:
Would you like to create a PR for this issue?

tom-thorne:
I can have a go - are there any docs, or please could you give me a hint as to what function to look at?

kamilmysliwiec:
That's the method you would have to modify: https://github.com/nestjs/swagger/blob/master/lib/plugin/visitors/controller-class.visitor.ts#L88 (after `extractedComments` if condition)

tom-thorne:
Thanks - and are there any unit tests I can leverage?

petrzjunior:
I would be too interested in supporting better instrospection. I usually write my comment in a TSDoc format with leading asterisks on each line, which are sadly not stripped.
``` typescript
/**
 * Public description, visible to customers.
 * @example
 * Summer Party
 */
```

[...]

PR #1207, opened 2021-02-15
feat(introspectComments): Enhanced introduction comments using tsdocs

juanmiguelbesada:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1097 

## What is the new behavior?
With this PR comments can be analysed with `@microsoft/tsdoc` parser to generate enhanced OpenApi spec based on comments rather than using Decorators.

In this PR only adds support for summary and description blocks, but we can keep improving it to add @examples, @params or whatever we need.

This PR keeps BC (while adding a package is not a bc). The new behaviour only works when setting `controllerKeyOfComment` to `null`.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/lib/plugin/utils/ast-utils.ts
+++ b/lib/plugin/utils/ast-utils.ts
@@ -18,6 +18,7 @@ import {
   TypeFormatFlags
 } from 'typescript';
 import { isDynamicallyAdded } from './plugin-utils';
+import { DocComment, DocExcerpt, DocNode, ParserContext, TSDocParser } from '@microsoft/tsdoc';
 
 export function isArray(type: Type) {
   const symbol = type.getSymbol();
@@ -100,6 +101,30 @@ export function getDefaultTypeFormatFlags(enclosingNode: Node) {
   if (enclosingNode && enclosingNode.kind === SyntaxKind.TypeAliasDeclaration)
     formatFlags |= TypeFormatFlags.InTypeAlias;
   return formatFlags;
+}
+
+export function getNodeDocs(
+  node: Node
+): DocComment {
+  const tsdocParser: TSDocParser = new TSDocParser();
+  const parserContext: ParserContext = tsdocParser.parseString(node.getFullText());
+  return parserContext.docComment;
+}
+
+export function docNodeToString(docNode: DocNode): string {
+  let result = '';
+
+  if (docNode) {
+    if (docNode instanceof DocExcerpt) {
+      result += docNode.content.toString();
+    }
+
+    for(const childNode of docNode.getChildNodes()) {
+      result += docNodeToString(childNode);
+    }
+  }
+
+  return result.trim();
 }
 
 export function getMainCommentAndExamplesOfNode(
--- a/lib/plugin/visitors/controller-class.visitor.ts
+++ b/lib/plugin/visitors/controller-class.visitor.ts
@@ -4,8 +4,9 @@ import { ApiOperation, ApiResponse } from '../../decorators';
 import { PluginOptions } from '../merge-options';
 import { OPENAPI_NAMESPACE } from '../plugin-constants';
 import {
+  docNodeToString,
   getDecoratorArguments,
-  getMainCommentAndExamplesOfNode
+  getMainCommentAndExamplesOfNode, getNodeDocs
 } from '../utils/ast-utils';
 import {
   getDecoratorOrUndefinedByNames,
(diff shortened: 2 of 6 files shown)


kamilmysliwiec:
Thanks for your contribution!

> With this PR comments can be analysed with @microsoft/tsdoc parser to generate enhanced OpenApi spec based on comments rather than using Decorators.

I'm trying to understand what benefits this change would bring us. I'm guessing the final goal would be to get rid of the current way we extract examples etc from the comment block (using regular expressions) and just use this package instead?

juanmiguelbesada:
> I'm trying to understand what benefits this change would bring us. I'm guessing the final goal would be to get rid of the current way we extract examples etc from the comment block (using regular expressions) and just use this package instead?

The main goal is to get rid of decorators as much as possible to provide doc info. Not just examples or summary.
At this moment for example, if you want to provide a description for a operation, you have to use @ApiOperation just for including a description, when it is something that, maybe, you already have done in your comments.

Example:


````ts
/**
 * This is what I spect to be the operation summary
 * 
 * @remarks
 * And we already have this operation really well documented in our comments, because we generate more documentation 
 * than just Swagger. AKA Compodoc
 */
@Get("/foo")
@ApiOperation({ description: "And we already have this operation really well documented in our comments, because we generate more documentation than just Swagger. AKA Compodoc" })
public fooAction() {}
````
vs
````ts
/**
 * This is what I spect to be the operation summary
 *
 * @remarks
 * And we already have this operation really well documented in our comments, because we generate more documentation 
 * than just Swagger. AKA Compodoc
 */
@Get("/foo")
public fooAction() {}
````

IMHO is more easy to use tsdoc than playing with regex

**Comment by exxocism** [comment]
YES 👍

**Comment by OscarMeier** [comment]
It would be really great if it also took `@throws` and `@summary` into account:

```
/**
 * Finds one cat.
 *
 * @summary Get one cat
 * 
 * @throws {BadRequestException} If wrong or missing parameters
 * @throws {NotFoundException} If not found
 *
 * @param {FindOneParamsDto} params
 * @param {LocaleQueryDto} query
 * @returns {Observable<Cat>} The corresponding cat
 */
```
would get parsed to:

```
@ApiOperation({
  description: 'Finds one cat.' // already implemented
  summary: 'Get one cat', // new
})
@ApiBadRequestResponse({
  description: 'If wrong or missing parameters', // new
})
@ApiNotFoundResponse({
  description: 'If not found', // new
})
// ApiOkResponse, ApiParam, ApiQuery and ApiBody get already generated by the swagger plugin.
@Get('cat/:id')
findOneCat(
  @Param() params: FindOneParamsDto,
  @Query() query: LocaleQueryDto,
): Observable<Cat> {}
```

thawankeane:
Any update of this?

I loved @OscarMeier suggestion, but apparently this PR is currently paused :(

Should we threat those features into a new issue (or PR)?

juanmiguelbesada:
@kamilmysliwiec Do you still find this interesting? I can back to work on this and added the features @OscarMeier mentioned...

kamilmysliwiec:
Sounds great @juanmiguelbesada!

kenerwin88:
I want this so bad! :'(

Uelb:
Looking at the code on this PR, it seems like it's ready for the first part of the feature, that is to accept the summary section as the api operation summary and the remarks section as api operation description ?
@juanmiguelbesada  is there something else you need to make it mergable ? I'd be glad to help if that is the case, as I would love to use this feature.

juanmiguelbesada:
> Looking at the code on this PR, it seems like it's ready for the first part of the feature, that is to accept the summary section as the api operation summary and the remarks section as api operation description ? @juanmiguelbesada is there something else you need to make it mergable ? I'd be glad to help if that is the case, as I would love to use this feature.

Sorry folks but couldn't get time to work on this yet.

@Uelb First thing will be to rebase master because this branch is pretty old. Then we could:
1. Merge and work in additional features in other PR
2. Implement [@OscarMeier suggested features](https://github.com/nestjs/swagger/pull/1207#issuecomment-1233949787)

My vote is to finish this up and then work in additional features in a different PR

Uelb:
@juanmiguelbesada Alright, rebasing without knowing the codebase can be tricky, would it be easier in this case to just restart and copy paste the code that is still interesting ? There is a huge number of different commits, I tried to perform the rebase but after 10 minutes, I think it's not worth it. What do you think ? Should I try and reproduce the first feature using part of your code (the summary part) ?

emretepedev:
I'm looking forward to this feature!

kamilmysliwiec:
Let's track this here https://github.com/nestjs/swagger/pull/2822

Issue #1212, opened 2021-02-18
Support for x-codegen-request-body-name


maxfriedmann:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

With @nestjs/swagger@3.x and OAS2, body parameters were named after their model names. After 18 months of tedious migration we finally made it to OAS3 and @nestjs/swagger@4.x. Now the body names are generated based on the class name, not the name of the body parameter anymore. This would lead to yet another migration round.

## Expected behavior

https://openapi-generator.tech/docs/swagger-codegen-migration/#body-parameter-name describes, how you can migrate your swagger/OAS json to provide a name again. I would like to be able to set the `x-codegen-request-body-name` via `@ApiBody()` as well.

## Minimal reproduction of the problem with instructions

```
  @Post()
  @ApiOperation({ summary: "Creates messages", operationId: "sendMessage" })
  @ApiCreatedResponse({ type: Message })
  public async sendMessage(@Body() message: CreateMessageDto): Promise<Message> {
    (...)
  }
```
This used to generate this method:
`sendMessage({message: CreateMessageDto})`
Now it is 
`sendMessage({createMessageDto: CreateMessageDto})` 
(via typescript-axios generator from https://openapi-generator.tech/docs/generators/typescript-axios)

## What is the motivation / use case for changing the behavior?

We're generating over 40 apis which are used by several projects. If we cannot find a solution for this issue, we would need to migrate all of them as well.

## Environment

<pre><code>
Nest version: 7.6.12

 
For Tooling issues:
- Node version: 14.15.0
- Platform:  Mac, Linux and Windows

Others:
We migrated from nestjs 6.3.x and nestjs/swagger 3.x yesterday, so now we're on the latest versions and facing this problem
</code></pre>

kamilmysliwiec:
As this is way too tool-specific, we don't plan to include & hardcode this extra metadata as part of the auto-generated OpenAPI document. If you need that, I'd suggest forking this package OR pulling its source code directly into your project's repository.

PR #1213, opened 2021-02-18
Add body name 

maxfriedmann:
... which will produce the x-codegen-request-body-name extension in the root of the document

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/nestjs/swagger/issues/1212

## What is the new behavior?
`@ApiBody({name: "Foo"}) is working (again) as expected

```
[ ] Yes
[x] No
```

## Other information
- For us, coming from @nestjs/swagger@3.x it is not a new feature, it used to work like this. Nevertheless, for 4.x it could be a feature? 
- I checked "Docs have been added" since the current documentation is valid: It does not state that, during the migration, you would have lost this feature

### DIFF
--- a/lib/swagger-explorer.ts
+++ b/lib/swagger-explorer.ts
@@ -353,6 +353,8 @@ export class SwaggerExplorer {
       ...omit(requestBody, keysToRemove),
       ...this.mimetypeContentWrapper.wrap(consumes, pick(requestBody, 'schema'))
     };
+    if (requestBody.name)
+      document.root["x-codegen-request-body-name"] = requestBody.name;
     return document;
   }
 
--- a/lib/decorators/api-body.decorator.ts
+++ b/lib/decorators/api-body.decorator.ts
@@ -19,6 +19,7 @@ interface ApiBodyMetadata extends RequestBodyOptions {
   type?: Type<unknown> | Function | [Function] | string;
   isArray?: boolean;
   enum?: SwaggerEnumType;
+  name?: string;
 }
 
 interface ApiBodySchemaHost extends RequestBodyOptions {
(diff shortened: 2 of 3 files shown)


lib/swagger-explorer.ts
```diff
@@ -353,6 +353,8 @@ export class SwaggerExplorer {
       ...omit(requestBody, keysToRemove),
       ...this.mimetypeContentWrapper.wrap(consumes, pick(requestBody, 'schema'))
     };
+    if (requestBody.name)
+      document.root["x-codegen-request-body-name"] = requestBody.name;
```
kamilmysliwiec, line 357: Why is this needed?

maxfriedmann: If the property is not set, the default value will be used (the class name). see https://openapi-generator.tech/docs/swagger-codegen-migration/#body-parameter-name. With this change and the `@ApiBody({name: "..."})` which we still have in our code from @nestjs/swagger@3.x, our generated APIs look exactly like before.

maxfriedmann: Can I provide any additional information @kamilmysliwiec to support you in your merge decision? 

I know that the generators from openapi-generator.tech aren't the only api generators out there and `x-codegen-request-body-name` seems to be a non-standard for now. But it seems that others are picking up that property as well, see https://github.com/swagger-api/swagger-codegen-generators/pull/827/files

kamilmysliwiec:
Thanks for your contribution! However, I can't merge this PR due to reasons outlined here https://github.com/nestjs/swagger/issues/1212#issuecomment-802841166

PR #1262, opened 2021-03-17
feat(config): merge passed schemas, links, ... in createDocument

### LINKED ISSUE #1244: Better support for raw component / schemas definitions
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

I have a complex model, basically a top-level oneOf, which is not supported by @nestjs/swagger:

```ts

/**
 * PersonGet is supposed to be CompanyGet | UserGet
 *
 * Nestjs swagger does not support a oneOf at the top level of a model. As such we create an OpenAPI definition manually
 */

type SchemaObject = OpenAPIObject['components']['schemas']['key']

export class PersonGet {}

export const personSchemaObject: SchemaObject = {
  oneOf: [
    {
      $ref: getSchemaPath(CompanyGet),
    },
    {
      $ref: getSchemaPath(UserGet),
    },
  ],
  discriminator: {
    propertyName: '_resolveType',
    mapping: {
      user: getSchemaPath(UserGet),
      company: getSchemaPath(CompanyGet),
    },
  },
}
```

I would like to be able to add this to the generated `OpenApiObject` in a simple manner

## Expected behavior

The current swagger implementations provides the following function:

```ts
export declare class SwaggerModule {
  static createDocument(
    app: INestApplication,
    config: Omit<OpenAPIObject, 'paths'>,
    options?: SwaggerDocumentOptions
  ): OpenAPIObject;
}
```
Ideally, a component should be able to be added to `config`:

```ts
config: {
  components: {
    schemas: {
       PersonGet: personSchemaObject
    }
  }
}
```

Unfortunately, the passed components are not merged deeply:

```ts
    // internal @nestjs/swagger code, in swagger-module.ts
    const document = swaggerScanner.scanApplication(app, options);
    document.components = {
      ...(config.components || {}),
      ...document.components
    };
```

I see two ways to address the problem:

- Change the behavior of `SwaggerModule.createDocument` to merge `config.components` one level deeper
- Create a new class decorator e.g. `@ApiRawDefinition` which contains the raw definition of the model, applied to the model class. 

## What is the motivation / use case for changing the behavior?

I want to create a model that is a `oneOf` of two other models. This is not supported yet, so I want to be able to specify the raw openapi definition somewhere.

Others:

I would be willing to make a PR, but I need to know which solution is preferrable :)

coyotte508:
Merge the passed config's components one level deeper, so as to be able to
specify individual schemas that will not be overwritten by the generated document.

Fix #1244

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1244 

## What is the new behavior?

`config.components` is merged one level deeper.

There is an alternative solution to #1244 which is to create a new decorator.

```
[x] Yes
[ ] No
```

Schemas, links, examples, responses, ... passed to `SwaggerModule.createDocument` via `config.components` are no longer overwritten if definitions of the same kind are present in the generated document.

They will be overwritten only if a definition of the same kind AND with the same name is present.

The amount of existing applications that would need to migrate should be relatively low, as the config will typically be handcrafted with nothing extra unwanted.

### DIFF
--- a/lib/document-builder.ts
+++ b/lib/document-builder.ts
@@ -173,7 +173,7 @@ export class DocumentBuilder {
     return this;
   }
 
-  public build(): Omit<OpenAPIObject, 'components' | 'paths'> {
+  public build(): Omit<OpenAPIObject, 'paths'> {
     return this.document;
   }
 }
--- a/e2e/validate-schema.e2e-spec.ts
+++ b/e2e/validate-schema.e2e-spec.ts
@@ -1,20 +1,30 @@
+import { INestApplication } from '@nestjs/common';
 import { NestFactory } from '@nestjs/core';
 import { writeFileSync } from 'fs';
 import { join } from 'path';
 import * as SwaggerParser from 'swagger-parser';
-import { DocumentBuilder, OpenAPIObject, SwaggerModule } from '../lib';
+import {
+  DocumentBuilder,
+  getSchemaPath,
+  OpenAPIObject,
+  SwaggerModule
+} from '../lib';
 import { ApplicationModule } from './src/app.module';
+import { Cat } from './src/cats/classes/cat.class';
+import { TagDto } from './src/cats/dto/tag.dto';
+import type { OpenAPIV3 } from 'openapi-types';
 
 describe('Validate OpenAPI schema', () => {
-  let document: OpenAPIObject;
+  let app: INestApplication;
+  let options: Omit<OpenAPIObject, 'paths'>;
 
   beforeEach(async () => {
-    const app = await NestFactory.create(ApplicationModule, {
+    app = await NestFactory.create(ApplicationModule, {
       logger: false
     });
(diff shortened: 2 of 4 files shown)


lib/document-builder.ts
```diff
@@ -173,7 +173,7 @@ export class DocumentBuilder {
     return this;
   }
 
-  public build(): Omit<OpenAPIObject, 'components' | 'paths'> {
+  public build(): Omit<OpenAPIObject, 'paths'> {
```
coyotte508, line 176: Same type as `this.document`

kamilmysliwiec:
LGTM

Issue #1106, opened 2020-12-22
swagger plugin generates code that throws Cannot find module '../../../stream'


tom-thorne:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

When a route returns a type that is defined in a core nodejs module, the distribution js that is created has an invalid require path.

For example

```
import { Readable } from 'stream';
...
    @Get(':id')
    async getExport(@Param('id') id: string): Promise<Readable> {
        return getExportFileStream(id);
    }   
```
generates:

```
__decorate([
    common_1.Get(':id'),
    openapi.ApiResponse({ status: 200, type: require("../../../stream").Readable }),  // <-- this is invalid
  ....
], EnrichmentImportExportController.prototype, "getExport", null);
```

## Expected behavior

It creates a valid path

```
__decorate([
    common_1.Get(':id'),
    openapi.ApiResponse({ status: 200, type: require("stream").Readable }), 
  ....
], EnrichmentImportExportController.prototype, "getExport", null);
```

## Environment

<pre><code>
Nest version: 7.5.1
"@nestjs/swagger": "4.7.6"
</code></pre>

kamilmysliwiec:
Please provide a minimum reproduction repository.

AlessandroMrli:
I had a similar issue
I needed to override the default by using this
	@ApiResponse({
		status: 200,
		type: Readable,
	})

kamilmysliwiec:
Fixed in the latest release

deviantfero:
I'm having this same issue with other libraries:
```
internal/modules/cjs/loader.js:797                                                                                                  
    throw err;                                                                                                                      
    ^                                                                                                                               
                                                                                                                                    
Error: Cannot find module '../../amazon-cognito-identity-js'                                                                        
```

```
__decorate([
    common_1.Post('refresh'),
    swagger_1.ApiExcludeEndpoint(),
    openapi.ApiResponse({ status: 201, type: require("../../amazon-cognito-identity-js").CognitoUserSession }),
    __param(0, common_1.Body()),
    __metadata("design:type", Function),
    __metadata("design:paramtypes", [refresh_token_dto_1.RefreshTokenDto]),
    __metadata("design:returntype", Promise)
], AuthController.prototype, "refreshToken", null);
```
[...]

PR #1318, opened 2021-04-28
fix(plugin): check if a importPath is a package

deviantfero:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When using a type provided by a node module the plugin would try to assemble a relative path that led to an error when trying to run the server, on release `4.8.0` an exception was added for the `stream` package, that fix worked only for that specific package, but other packages still were throwing out the same error (refer to issue #1106) 

Issue Number: #1106 

## What is the new behavior?
Using `require.resolve(name: string)` we can check if a string is a package, this function throws an error if the package is not found in the project's installed packages, we can catch this error and default to the usual behavior that works for all other files in the project. 

```
[ ] Yes
[x] No
```

## Other information
node's `require.resolve(name:string)` documentation:
https://nodejs.org/api/modules.html#modules_require_resolve_request_options

### DIFF
--- a/lib/plugin/utils/plugin-utils.ts
+++ b/lib/plugin/utils/plugin-utils.ts
@@ -111,8 +111,6 @@ export function hasPropertyKey(
     .some((item) => item.name.getText() === key);
 }
 
-const NATIVE_NODEJS_LIBRARIES = 'stream';
-
 export function replaceImportPath(typeReference: string, fileName: string) {
   if (!typeReference.includes('import')) {
     return typeReference;
@@ -124,9 +122,10 @@ export function replaceImportPath(typeReference: string, fileName: string) {
   importPath = convertPath(importPath);
   importPath = importPath.slice(2, importPath.length - 1);
 
-  if (NATIVE_NODEJS_LIBRARIES.includes(importPath)) {
+  try {
+    require.resolve(importPath)
     return typeReference.replace('import', 'require');
-  } else {
+  } catch (_error) {
     let relativePath = posix.relative(posix.dirname(fileName), importPath);
     relativePath = relativePath[0] !== '.' ? './' + relativePath : relativePath;
 


kamilmysliwiec:
I'm not sure if I'm following. We do want to support packages that are non-native. For example, imaging having a private package that is published under your company's NPM scope where you keep certain common types that you share between apps. In this case, we definitely don't want to simply ignore them when referred from an app-specific DTO class. The intention here isn't to ignore **all** node packages.

deviantfero:
I don't think we're ignoring any packages at all, unless I'm misunderstanding the use of `require.resolve`, the idea is to simply use the package name in the `require(...)` statement generated by the plugin instead of a relative path that is usually generated, this is a simple check to verify if the import is actually a package that `require` can use as is and avoid creating a relative path for that import if it is, for example, I'm using a type provided by the package `amazon-cognito-identity-js` as a return type in one of my endpoints, currently the plugin produces this:

```
__decorate([
    common_1.Post('refresh'),
    swagger_1.ApiExcludeEndpoint(),
    openapi.ApiResponse({ status: 201, type: require("../../amazon-cognito-identity-js").CognitoUserSession }),
    __param(0, common_1.Body()),
    __metadata("design:type", Function),
    __metadata("design:paramtypes", [refresh_token_dto_1.RefreshTokenDto]),
    __metadata("design:returntype", Promise)
], AuthController.prototype, "refreshToken", null);
```

which will fail inevitably when running the server, there is no such thing as `../../amazon-cognito-identity-js`, only `amazon-cognito-identity-js` we can check for this with `require.resolve('amazon-cognito-identity-js')`, if that call doesn't throw an exception then it means it's fine to just require it without the need for a relative path, with this fix the plugin would produce the appropriate:

```
__decorate([
    common_1.Post('refresh'),
    openapi.ApiResponse({ status: 201, type: require("amazon-cognito-identity-js").CognitoUserSession }),
    __param(0, common_1.Body()),
    __metadata("design:type", Function),
    __metadata("design:paramtypes", [refresh_token_dto_1.RefreshTokenDto]),
    __metadata("design:returntype", Promise)
], AuthController.prototype, "refreshToken", null);
```

kamilmysliwiec:
Oh alright, I think I've misunderstood! That makes a lot of sense, I'll review asap

deviantfero:
Thanks, if there's anything else you need me to do to get this merged let me know and I'll do it when I'm able to, I hope this can be helpful to others using this plugin, it has saved me a lot of typing for documenting properties :smile:

boy51:
I'm facing this exact error with `amazon-cognito-identity-js` so it would be awesome if this made it 👍

deviantfero:
@boy51 as a temporary solution I've opted to return `any` and loose that information in this endpoint for now, this is because it's so convenient to have the rest of the endpoints automatically documented, so I just deal with this one endpoint not being documented for now:

```
  @Post('refresh')
  async refreshToken(@Body() tokens: RefreshTokenDto) {
    try {
      return (await this.authService.refreshTokens(tokens)) as any;
    } catch (error) {
      throw new BadRequestException(error.message);
    }
  }
```

boy51:
Luckily I have some time before I need the docs, so I'm ok for now. 

If it doesnt make it until then I think its gonna be easy enough to set up a patch with patch-package.

kamilmysliwiec:
LGTM

Kolobok12309:
@deviantfero 
Big thanks for this PR, in v4.8.1 paths are broked and i can't build in CI(heroku)

deviantfero:
No problem, nest has been a blast to use, and the swagger plugin is really helpful, so I'm glad to contribute, glad it got merged

PR #1371, opened 2021-05-27
feature(decorators): Add ApiPaymentRequiredResponse decorator

linkRace:
## PR Checklist

- [X] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md

There are no relevant tests/docs I could see around these decorators, but if there are and I missed them, please point me in the correct direction.

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

To explicitly support [ HTTP Code 402](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/402) instead of needing to fall back to the default response, add this decorator

Issue Number: N/A

## What is the new behavior?
Can use the new decorator in a controller to define the expected error

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/lib/decorators/api-response.decorator.ts
+++ b/lib/decorators/api-response.decorator.ts
@@ -231,6 +231,12 @@ export const ApiUnsupportedMediaTypeResponse = (
     status: HttpStatus.UNSUPPORTED_MEDIA_TYPE
   });
 
+export const ApiPaymentRequiredResponse = (options: ApiResponseOptions = {}) =>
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Thanks for your contribution! However, this has been discussed in the past and we decided to not add more "wrapper" decorators to the library as its API is already getting too bloated.

Issue #1204, opened 2021-02-14
@ApiQuery with enum has a different definition if used in combination with @Query


arkraft:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Currently, when defining an enum as a query parameter (`@ApiQuery({ name: "enumName", enum: SomeEnum, enumName: "SomeEnum" })`), there is a difference how the enum will be described in the OpenAPI Docs:
- When the enum is also used as a method parameter using the `@Query()` decorator, it will be inlined as a string type
- When the enum is not used as a method parameter, it will not be inlined but instead defined as a reference

The behavior changed from version 4.7.2 -> 4.7.3. Before, it was always defined as a reference. 

## Expected behavior
It should always be described as a reference

## Minimal reproduction of the problem with instructions
There is a test case in the `swagger-explorer.spec.ts` file using the `BarController` which has multiple Enums. None of them are used as method parameters and the test case works as expected. I forked the repo and defined the method parameter in the `BarController`.
- This is the repo: https://github.com/arkraft/swagger
- This is the failing Circle CI test: https://app.circleci.com/pipelines/github/arkraft/swagger/3/workflows/1c9128f7-f5a2-43aa-a001-d1a1feec5160/jobs/6

Additionally there is a reproduction repo with two methods. One does not defined a `@Query` parameter, the other one does. The `swagger.json` in the root folder shows the difference between the two:

https://github.com/arkraft/reproduction-nestjs-swagger-1204

Switching the version from 4.7.13 to 4.7.2 will show that bot enums are defined equally in the `swagger.json`

## What is the motivation / use case for changing the behavior?
We use `swagger-axios-codegen` for creating the API clients. With this generator, the enum is only extracted as an enum if it is defined as a reference to a separate schema. 

## Environment

<pre><code>
Nest version: 4.7.2 and 4.7.13

 
For Tooling issues:
- Node version: 14.7.0  
- Platform: Linux 

Others:

</code></pre>

kamilmysliwiec:
Would you like to create a PR for this?

arkraft:
I will see what i can do

kamilmysliwiec:
Closing due to inactivity

arkraft:
Sorry for the delay, but i opened a PR for this

[...]

PR #1375, opened 2021-05-30
fix(@nestjs/swagger): remove explicitly defined params

arkraft:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When defining enums with `@ApiQuery` or `@ApiParm`, their type will only be defined as `$ref` if they are not added as parameters of the method. If they are added to the method with `@Query` or `@Param`, the types will be overwritten. This behavior changed with the update 4.7.2 -> 4.7.3. Therefore i submit this as a bug fix. 

Issue Number: #1204 

## What is the new behavior?

Whether the enum is additionally defined as `@Param` or `@Query`, the result will be the same and the enum will be a reference object.

```
[x] Yes
[ ] No
```
```
@ApiQuery({ enumName: "QueryEnum", name: "test", enum: QueryEnum })
postHere(@Query("test") test: QueryEnum) {}
```

currently looks like this:

```
{
  in: "query",
  name: "test",
  schema: {
    type: "string",
    enum: [...]
  }
  ...
}
```

and it now will look like this:

```
{
  in: "query",
  name: "test",
  schema: {
    $ref: ".../QueryEnum"
  }
  ...
}
```

The second example is also the result if the method looks like this:

```
@ApiQuery({ enumName: "QueryEnum", name: "test", enum: QueryEnum })
postHere() {}
```


jmcdo29 (approved):
Lgtm. Reviewed from my phone so I may have missed something

iamdenny:
when is this going be merged?

arkraft:
I did not have much time to work on it lately. Now i kinda fucked up during merging and it shows that there are many changes which is not the case. Therefore i started a new PR with the fix and need to fix a test case now, hopefully this new PR will work better

arkraft:
> when is this going be merged?

@iamdenny I created a new PR with the changes which is cleaner and i also found out, why the test was failing in the pipeline. Here is the updated PR: #1733

PR #1480, opened 2021-08-02
feat: Add default type to Api Error Responses

dariusj18:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently each of the error api response decorators do not specify a type, despite the responses having common information. 

## What is the new behavior?

This PR adds a DTO class for each response type and sets them as the default value for `type`.  It is set before the spread operator so that the use can still pass in their own type.

```
[x] Yes
[ ] No
```

This adds a default type to each response, which may not be expected.

### DIFF
--- a/lib/decorators/api-response.decorator.ts
+++ b/lib/decorators/api-response.decorator.ts
@@ -9,6 +9,8 @@ import {
 } from '../interfaces/open-api-spec.interface';
 import { getTypeIsArrayTuple } from './helpers';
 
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
This will generate an incorrect schema for everyone with a custom global exception filter. 

Let's track the alternative solution/approach to this issue here https://github.com/nestjs/swagger/issues/2548

Issue #1037, opened 2020-11-10
Provide a shim file


kamilmysliwiec:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Expected behavior

Something similar to what `@nestjs/graphql` provides.

More on that topic https://github.com/nestjs/graphql/issues/1082
Example implementation: https://github.com/nestjs/graphql/pull/1184/files

## Environment

<pre><code>
Nest version: X.Y.Z

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

mkhraisha:
are there any updates on this?

nick-lehmann:
@kamilmysliwiec would you accept a merge request for this? We have created a small shim file for this package at work that includes all the exports we need. I guess I could start from there, add the remaining exports and submit a pull request later this week.

jmcdo29:
@nick-lehmann that sounds great! Thanks for the help

nick-lehmann:
Is there a reason why the [shim file](https://github.com/maayanbar13/graphql/blob/e98f3ec6c78c5ff9a8ebefc503c67d99ab4f5bc2/lib/extra/graphql-model-shim.ts) for `@nestjs/graphql` is written in typescript, unlike the [typeorm shim](https://github.com/typeorm/typeorm/blob/master/extra/typeorm-model-shim.js) plain javascript? As far as I can tell, the arguments for the decorators do not need any type information at build time and it would make the shim file quite a lot easier. So instead of 

```typescript
# shim.ts
export function Directive(sdl: string,
    ): MethodDecorator & PropertyDecorator & ClassDecorator {
    return (target: Function | Object, key?: string | symbol) => {}
}
```

one could simply do

```javascript
# shim.js
function Directive() {
    return () => {}
}
exports.Directive = Directive
```

The second way is how I have done it so far and it is working fine for us.

kamilmysliwiec:
Either approach should be fine. What issues did you run into when using `ts` files?

nick-lehmann:
Sorry for the late reply. I added the shim file and created a pull request for it.

No, I did not experience any issues with the typescript file. I just wondered if there might be any downside that I did not think of.

kamilmysliwiec:
[...]

PR #1505, opened 2021-08-17
feat(): add shim file for browsers

nick-lehmann:
## PR Checklist
Please check if your PR fulfils the following requirements:

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1037

## What is the new behavior?

Developers can use classes decorated with decorators from `@nestjs/swagger` in typescript applications that are run in the browser. This is done by telling either the typescript compiler or a bundler used in the browser application to use the newly added shim file instead of the actual package. The shim file is a lot smaller in size and does not introduce any dependencies on node packages that would otherwise require even larger polyfills.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/lib/extra/swagger-shim.ts
+++ b/lib/extra/swagger-shim.ts
@@ -1,0 +1,166 @@
+/* eslint-disable @typescript-eslint/no-empty-function */
+export function ApiProperty() {
+  return () => {};
+}
+export function ApiPropertyOptional() {
+  return () => {};
+}
+export function ApiResponseProperty() {
+  return () => {};
+}
+export function ApiBasicAuth() {
+  return () => {};
+}
+export function ApiBearerAuth() {
+  return () => {};
+}
+export function ApiBody() {
+  return () => {};
+}
+export function ApiConsumes() {
+  return () => {};
+}
+export function ApiCookieAuth() {
+  return () => {};
+}
+export function ApiExcludeEndpoint() {
+  return () => {};
+}
+export function ApiExcludeController() {
+  return () => {};
+}
+export function ApiExtraModels() {
+  return () => {};
+}
+export function ApiHeader() {
+  return () => {};
+}
+export function ApiHeaders() {
+  return () => {};
+}
+export function ApiHideProperty() {
+  return () => {};
+}
+export function ApiOAuth2() {
+  return () => {};
+}
+export function ApiOperation() {
+  return () => {};
+}
+export function ApiParam() {
+  return () => {};
+}
+export function ApiProduces() {
+  return () => {};
+}
+export function ApiQuery() {
+  return () => {};
+}
+export function ApiResponse() {
+  return () => {};
+}
+export function ApiOkResponse() {
+  return () => {};
+}
+export function ApiCreatedResponse() {
+  return () => {};
+}
+export function ApiAcceptedResponse() {
+  return () => {};
+}
+export function ApiNoContentResponse() {
+  return () => {};
+}
+export function ApiMovedPermanentlyResponse() {
+  return () => {};
+}
+export function ApiFoundResponse() {
+  return () => {};
+}
+export function ApiBadRequestResponse() {
+  return () => {};
+}
+export function ApiUnauthorizedResponse() {
+  return () => {};
+}
+export function ApiTooManyRequestsResponse() {
+  return () => {};
+}
+export function ApiNotFoundResponse() {
+  return () => {};
+}
+export function ApiInternalServerErrorResponse() {
+  return () => {};
+}
+export function ApiBadGatewayResponse() {
+  return () => {};
+}
+export function ApiConflictResponse() {
+  return () => {};
+}
+export function ApiForbiddenResponse() {
+  return () => {};
+}
+export function ApiGatewayTimeoutResponse() {
+  return () => {};
+}
+export function ApiGoneResponse() {
+  return () => {};
+}
+export function ApiMethodNotAllowedResponse() {
+  return () => {};
+}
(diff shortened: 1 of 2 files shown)


nick-lehmann:
If it makes sense, I could also try to add a test that checks if the shim file includes all types exported from the package. This would catch the case where somebody adds a test and accidentally forgets to add it to the shim file.

kamilmysliwiec:
@nick-lehmann sounds great!

nick-lehmann:
@kamilmysliwiec Done!

danielglazer:
I am having the same problem with classes decorated with decorators from @nestjs/swagger that I want to publish as a package to be consume from both server & client. @nick-lehmann , after your solution would be merged. can you please explain or point me in the right direction on how to apply it?

kamilmysliwiec:
LGTM

ekselys:
Great solution @nick-lehmann .
By the way, OmitType, PickType, and so on will loose their ability to create classes from existing ones.
It would be helpful if these classes continue to work as is, because it's very common to create a DTO by removing some properties from a model (ex: CreateUserDTO will have all User properties without the ID).

Is there anyway to decouple the decorators and helper types from the rest of the library?

luskin:
@nick-lehmann Would you be able to give an example of how to use this for shared types between a NestJS backend and a React frontend?

javifm86:
> 
> is required to share types between API and Web
>       '@nestjs/swagger': path.resolve(__dirname, 'n

Hey @luskin, thank you for pointing me in the right direction, I am trying to replicate something similar. Could you please elaborate and show your solution in detail? It would be interesting for me.

This is my approach:

I have a common dto in a `common` folder, that I consume installing as dependency in both projects, React for frontend and NestJS for backend. Example:

```ts
import { ApiProperty } from '@nestjs/swagger';

export class TestDto {
  @ApiProperty({
    description: 'Property description',
  })
  test: string;
}
```

In React I can use it this way, and the type works properly, I get autocompletion:
```tsx
import { TestDto } from "app-common";
 const a : TestDto = {
    test: '',
  };
```
And I can use the same one in NestJS.

But I am uncertain if this is correct and the swagger-shim is coming into play or not 😅.

Thanks in advance. Kind regards.

luskin:
> > is required to share types between API and Web
> > '@nestjs/swagger': path.resolve(__dirname, 'n
> 
> Hey @luskin, thank you for pointing me in the right direction, I am trying to replicate something similar. Could you please elaborate and show your solution in detail? It would be interesting for me.
> 
> This is my approach:
> 
> I have a common dto in a `common` folder, that I consume installing as dependency in both projects, React for frontend and NestJS for backend. Example:
> 
> ```ts
> import { ApiProperty } from '@nestjs/swagger';
> 
> export class TestDto {
>   @ApiProperty({
>     description: 'Property description',
>   })
>   test: string;
> }
> ```
> 
> In React I can use it this way, and the type works properly, I get autocompletion:
> 
> ```tsx
> import { TestDto } from "app-common";
>  const a : TestDto = {
>     test: '',
>   };
> ```
> 
> And I can use the same one in NestJS.
> 
> But I am uncertain if this is correct and the swagger-shim is coming into play or not 😅.
> 
> Thanks in advance. Kind regards.

@javifm86 You need to point '@nestjs/swagger' in your webpack settings to the provided js shim file. Here is an example craco config:

```js
/* eslint-disable @typescript-eslint/no-var-requires */
/* eslint-disable no-undef */
const path = require('path')

module.exports = {
  module: {
    rules: [
      {
        test: /\.worker\.js$/,
        use: { loader: 'worker-loader' }, // Mapbox GL
      },
    ],
  },
  babel: {
    loaderOptions: {
      ignore: ['../../node_modules/mapbox-gl/dist/mapbox-gl.js'], // Mapbox GL
    },
  },
  webpack: {
    alias: {
      // This is required to share types between API and Web
      '@nestjs/swagger': path.resolve(__dirname, '../../node_modules/@nestjs/swagger/dist/extra/swagger-shim.js'),
    },
  },
}

```

javifm86:
Thanks for the reply @luskin ! I have a monorepo, I think I will create a common folder with the dtos and I will copy with a script or tool the source dtos files to each project to be consumed. And with that configuration in the frontend it will be working.

krzysztofMlczk:
@kamilmysliwiec does it mean that in order to shim those decorators out (so I can share types between BE and FE) I still need `nestjs/swagger` as a dependency in my frontend?

Also when I try to use mapped types in those shared types I end up with bundling issues in my frontend app.

As @ekselys mentioned [in his comment](https://github.com/nestjs/swagger/pull/1505#issuecomment-934811425) won't using shim file break the functionality of mapped types?

krzysztofMlczk:
> I wrote a post in my blog with the approach I followed, just in case it is useful for someone: [Share Datos between Nest and React](https://javifm.com/en/blog/share-dtos-between-nestjs-and-react/)

@javifm86 but have you used mapped types in the types that you wanted to share between backend and frontend? If so how did you manage to bundle your frontend app? E.g. `OmitType` depends on `@nestjs/common` which breaks bundling in my case.

maciejgoscinski:
@krzysztofMlczk @kamilmysliwiec

**Mapped types do not work in the browser with this shim**.

In the shim:
```
exports.PartialType = PartialType;
function PickType() {
    return () => { };
}
```

Usage as per docs:
```
export class UpdateCatDto extends PartialType(CreateCatDto) {}
```

Results in
```
Uncaught TypeError: Class extends value () => {} is not a constructor or null
```

Issue #1653, opened 2021-11-02
First-level schema type in @ApiHeader is ignored or overwritten to string


simonkurka:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

```typescript
@ApiHeader({ name: 'x-custom', schema: { type: 'array', items: { type: 'string' } } })
```
Results in `string[string]` OpenAPI specification.

### Minimum reproduction code

https://github.com/simonkurka/nestjs-apiheader-mwe

### Steps to reproduce

1. `npm i`
2. `npm start:dev`
3. See error at http://localhost:3000/api/#/default/AppController_getHello

### Expected behavior

It should result in `array[string]` in OpenAPI specification.

### Package version

5.1.4

### NestJS version

8.1.2

### Node.js version

16.11.1

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

simonkurka:
Setting the parameter type to `Array<string>` changes the behaviour, but this is wrong because of https://github.com/nestjs/nest/issues/8175

kamilmysliwiec:
Would you like to create a PR for this issue?

simonkurka:
At least I can try.
Would you check that I am following the guidelines and that there are no unwanted side effects if possible?

PR #1655, opened 2021-11-03
fix(swagger): @ApiHeader default type instead of overwriting

simonkurka:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1653

## What is the new behavior?
string ist the default type for ApiHeader, but can be overwritten.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/lib/decorators/api-header.decorator.ts
+++ b/lib/decorators/api-header.decorator.ts
@@ -26,8 +26,8 @@ export function ApiHeader(
       description: options.description,
       required: options.required,
       schema: {
-        ...(options.schema || {}),
-        type: 'string'
+        type: 'string',
+        ...(options.schema || {})
       }
     },
     negate(isUndefined)


simonkurka:
I think there's no docs update needed.
A test-case might be a good idea. Do you have a hint where this should reside in?
Is the commit message ok? swagger is not in the [supported scopes](https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md#commit) but this seemed to be meaningful to me.

kamilmysliwiec:
Can you explain what other types we should support? I though header is always a string anyway

simonkurka:
https://swagger.io/docs/specification/data-models/data-types/ lists possible types in swagger.

The case here is that the swagger type at first level is always overwritten with string, which is actually wrong as swagger supports more types.

This has nothing to do with the fact that HTTP-Headers are always strings. The case is the swagger output and as a result the swagger-ui generation which both are typed.

kamilmysliwiec:
Can you update e2e tests with non-string headers then?

yanavery:
@kamilmysliwiec, @simonkurka, what happened with this? Sorry to comment on a closed PR ...

We have a similar use case, although simpler, we need our header values to be of type `number` (not `string`) as they are used for data paging such as `X-Paging-Offset` and `X-Paging-Size`, but the Swagger file gets generated as follows:
```
...
"someGetMethod","parameters": [
 {
   "name":"X-Paging-Offset",
   "in":"header",
   "description":"The offset to start returning data from, starting at 0.",
   "required":false,
   "schema":{"type":"string"}
 },
 {
   "name":"X-Paging-Size",
   "in":"header",
   "description":"The number of records to return.",
   "required":false,
   "schema":{"type":"string"}
 }
],
...
```
Even though we have the following in our code (see `schema` -> `type` -> `number`):
```
// ====================================================================================================================
export function DataPaging() {
  return applyDecorators(
    ApiHeader({
      name: HEADER_DATA_PAGING_OFFSET,
      required: false,
      description: "The offset to start returning data from, starting at 0.",
      schema: { type: "number" },
    }),
    ApiHeader({
      name: HEADER_DATA_PAGING_PAGE_SIZE,
      required: false,
      description: "The number of records to return.",
      schema: { type: "number" },
    })
  );
}
```

The `"schema":{"type":"number"}` is simply ignored and it looks like the underlying code is just blindly putting `schema: { type: "string" },` which explains why this is happening ...

See https://github.com/nestjs/swagger/blob/b465c9a75dcf14b469cb420f6a6d3428d4f8e67b/lib/decorators/api-header.decorator.ts#L30

We use the generated Swagger file to then generate our Axios API client, which would be much better if properly typed instead of typed with `string` parameter values - should be `number` values:
```
/* 
 * @param {string} [xPagingOffset] The offset to start returning data from, starting at 0.
 * @param {string} [xPagingSize] The number of records to return.
 * @param {*} [options] Override http request option.
 * @throws {RequiredError}
 */
  someGetMethod: async (xPagingOffset?: string, xPagingSize?: string, options: AxiosRequestConfig = {}): 
  Promise<RequestArgs> => {
     const localVarPath = `/v1/someGetMethod`;
     ...
```

Any way this could be worked around or fixed?

Thanks!

cityreader:
I can confirm this PR has fix my issue when I want to set type to ApiHeader schema type to `boolean`

```
  @ApiHeader({
    name: 'Add-Components',
    description: 'Specifies if the response should contain the address broken down into its components.',
    schema: {
      type: 'boolean',
      default: false,
    },
  })
```

Before

Screenshot 2023-07-13 at 3 08 53 pm

After

Screenshot 2023-07-13 at 3 07 33 pm

Issue #1695, opened 2021-11-24
Parse @param from comments blocks when using the CLI plugin


adaliszk:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

As a Developer, when I'm using the compiler plugin, I wish that it would read the `@param` JsDoc/TsDoc from the code blocks of methods so that I would not require to add an `@ApiParam` decorator just to describe a parameter for the generated API specification.

The [documentation](https://docs.nestjs.com/openapi/cli-plugin) briefly touches on the capabilities of the comment block, mentioning that only description and examples work at the moment. There is also a related Automatically detect optional Query (#30) that should be considered when implementing this.

### Describe the solution you'd like

when using `introspectComments: true` with the code:
```typescript
/**
 * This comment is already working as the "description" field
 *
 * @param limit The amount of items to be returned
 */
@Get()
myMethod(@Query('limit') limit: number) {}
```

the generated Open API specification would fill out the description for that param automatically.

### Teachability, documentation, adoption, migration strategy

The documentation is pretty straitforward at the moment, I think an addition with a table to list all the supported JsDoc/TsDoc annotations would be a nice. With that, a link to raise a new feature request for more annotations could be added.

The migration stategy is a simple replacement:
```typescript
  /**
   * This comment is already working as the "description" field
   *
+  * @param limit The amount of items to be returned
   */
  @Get()
- @ApiParam({name: 'limit', description: 'The amount of items to be returned'})
  myMethod(@Query('limit') limit: number) {}
```

This JsDoc/TsDoc annotation usually already generated by most IDEs, so for the most part it would be only removal of the decorator.

### What is the motivation / use case for changing the behavior?

Most IDEs already auto-generate the `@param` JsDoc/TsDoc and they keep it up to date while the custom decorator is not maintained automatically. Furthermore, the annotation is shorter, and its purpose is to serve as the documentation description that is used by many tools outside of Nest JS.

For Code-Quality use case,  
this would mean less bloat for filling out descriptions, thus easier overviewability of the code. When a decorator is needed for custom behaviour, then it becomes much easier to spot as all decorator usages will be the advanced cases.

For Tool Integrations use case, such as documentation or specification generators,  
this would mean that the CLI plugin alignes with third parties automatically, as there would be a single source of truth for at least the description. Some custom behaviours may still differ, but those are special cases usually.

kamilmysliwiec:
[...]

PR #1720, opened 2021-12-12
feat(): add support for additional TSDoc tags

andrewda:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/swagger/issues/1695

Currently, nestjs/swagger only supports `@example` TSDoc tags.

## What is the new behavior?

This PR adds support for the [`@deprecated`](https://tsdoc.org/pages/tags/deprecated/) TSDoc tag, in addition to the existing support for [`@example`](https://tsdoc.org/pages/tags/example/) tag. It also defines a more solid interface for adding additional tag support in the future, such as support for [`@param`](https://tsdoc.org/pages/tags/param/) as discussed in #1695. 

Note that this PR doesn't yet add support for `@param`, since there likely has to be a larger discussion about how it should be implemented (e.g. what happens when the symbol name differs from the property name, like `@Query('pageLimit') limit: number`).

- [ ] Yes
- [x] No

## Other information

I tried to build this in a way that could be easily extendable in the future for additional tags. If anyone has suggestions on how this could be improve, please let me know!


joaocarmo:
@andrewda, just to add to @cqravinder's point, is there anything missing from this PR that we could help out with? It would be great if we could push this one through.

We're using this to generate an API Client and, right now, we have to resort to manual monkey patching to mark certain endpoints as `@deprecated` and inform our users.

andrewda:
Hey @joaocarmo, I went ahead and fixed the conflicts here. I believe this is still ready to go, just waiting for a review from the NestJS team!

kamilmysliwiec:
lgtm

its-dibo:
`@param` still not supported.


> Note that this PR doesn't yet add support for `@param`, since there likely has to be a larger discussion about how it should be implemented (e.g. what happens when the symbol name differs from the property name, like `@Query('pageLimit') limit: number`).

it should use the name that appear in swagger docs, because we are documenting the API, not the method itself, in contrast to the service methods which we document the method not the API routes

other suggestions:
- at least enable it when both names are identical
- allow documenting both `pageLimit` and `limit`, pick the right one for swagger's annotations i.e. `pageLimit`
- add additional optional param for swagger `@apiQuery pageLimit .....`

PR #1733, opened 2021-12-16
fix(@nestjs/swagger): remove parameters which are explicitly defined

arkraft:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When defining enums with `@ApiQuery` or `@ApiParm`, their type will only be defined as `$ref` if they are not added as parameters of the method. If they are added to the method with `@Query` or `@Param`, the types will be overwritten. This behavior changed with the update 4.7.2 -> 4.7.3. Therefore i submit this as a bug fix. 

Issue Number: #1204 

## What is the new behavior?
Whether the enum is additionally defined as `@Param` or `@Query`, the result will be the same and the enum will be a reference object.

```
[ ] Yes
[X] No
```
```
@ApiQuery({ enumName: "QueryEnum", name: "test", enum: QueryEnum })
postHere(@Query("test") test: QueryEnum) {}
```

currently looks like this:

```
{
  in: "query",
  name: "test",
  schema: {
    type: "string",
    enum: [...]
  }
  ...
}
```

and it now will look like this:

```
{
  in: "query",
  name: "test",
  schema: {
    $ref: ".../QueryEnum"
  }
  ...
}
```

The second example is also the result if the method looks like this:

```
@ApiQuery({ enumName: "QueryEnum", name: "test", enum: QueryEnum })
postHere() {}
```

I believe the new result to be the correct one. It was also the result in Versions prior to 4.7.3. I don't think its's breaking as the new result is the expected one.


e2e/validate-schema.e2e-spec.ts
```diff
@@ -41,6 +43,8 @@ describe('Validate OpenAPI schema', () => {
       .addSecurityRequirements('bearer')
       .addSecurityRequirements({ basic: [], cookie: [] })
       .build();
+
+    Reflect.deleteMetadata("swagger/apiParameters", CatsController.prototype.getWithEnumParam);
```
arkraft, line 46: I don't know if there is a cleaner way to do this. The problem was, that on the second test case the metadata was merged and it was not the expected schema containing the enum values but the ref path to the enum schema which did not exist at that time. So the schema which was added on the second test was corrupt

funkenstrahlen:
Can we get this merged? This is a really important PR for us.

lib/explorers/api-parameters.explorer.ts
```diff
@@ -83,3 +87,25 @@ function removeBodyMetadataIfExplicitExists(
...
+    properties: ParamWithTypeMetadata[],
+    explicitParams: any[]
+) {
+
+  const reducer = (a, p) => {
+    if (['query', 'path'].includes(p.in)) {
+      a.push(p.name);
+    }
+    return a;
+  }
```
micalevisk, line 101: I guess we can rename these parameters

- `a` -> `reflectedParams`
- `p` -> `paramWithTypeMetadata`

lib/explorers/api-parameters.explorer.ts
```diff
@@ -83,3 +87,25 @@ function removeBodyMetadataIfExplicitExists(
...
+
+  const reflectedParams = isEmpty(properties) ? [] : properties.reduce(reducer, []);
+  const explicitlyDefinedParams = isEmpty(explicitParams) ? [] : explicitParams.reduce(reducer, []);
+  const intersect = intersection(explicitlyDefinedParams, reflectedParams);
+
+  return omitBy(
+      properties,
+      (p) => intersect.includes(p.name),
```
micalevisk, line 109:
```suggestion
      (paramWithTypeMetadata) => intersect.includes(paramWithTypeMetadata.name),
```

lib/explorers/api-parameters.explorer.ts
```diff
@@ -83,3 +87,25 @@ function removeBodyMetadataIfExplicitExists(
   }
   return properties;
 }
+
+function removeParamMetadataIfExplicitExists(
+    properties: ParamWithTypeMetadata[],
+    explicitParams: any[]
+) {
+
```
micalevisk, line 95:
```suggestion
```

micalevisk (approved):

lib/explorers/api-parameters.explorer.ts
arkraft: Thank you for the review. I updated the variable names

funkenstrahlen:
@kamilmysliwiec Can you take a look at this and merge? ❤️ Unfortunately this is blocking us for a long time now.

arkraft:
Would it be possible to review this? I check the current versions of @nestjs/swagger all the time and none of them fix the problem. All our enums are broken in newer releases, so we still stick with 4.7.2, which is 3 major versions behind now. And with that we cannot update nestsjs to a newer version because of type mismatches. Please consider reviewing this PR.

arkraft:
Sorry, the type mismatches are not due to @nestjs/swagger. I tried it again and it was typescript 4.8.2 which causes problems with other libraries.

Bassadin:
Is it possible to merge this? It'd be really nice for our project.

micalevisk (approved):

arkraft:
I believe this is obsolete now.

PR #1873, opened 2022-03-31
fix: allof client generation handling

gausnes:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

NestJS swagger currently uses `allOf` in somewhat of a weird usage based on examples within the documentation. This is fine for rendering the specification, but causes issues when attempting to generate models and clients off of the emitted specification. [OpenAPI Generator](https://github.com/OpenAPITools/openapi-generator) is not inspect the `required` flag for composite schemas.

This is not the only solution to this issue, but was the simplest. I spent some time evaluating fixing this within the generator code, but it was significantly more involved. This change does not feel like an explicitly inerrant as the resulting object of compose schema use is an object.  

```yaml
properties:
  id:
    description: Identifier
    allOf:
      - $ref: '#/components/schemas/Identifier'
  name:
    type: string
required:
  - eventId
```

## What is the new behavior?

Typing these properties as `objects` helps the generator inspect the expected fields for correct generation. 

```yaml
properties:
  id:
    description: Identifier
    type: object
    allOf:
      - $ref: '#/components/schemas/Identifier'
  name:
    type: string
required:
  - eventId
```

- [ ] Yes
- [x] No

### DIFF
--- a/lib/services/schema-object-factory.ts
+++ b/lib/services/schema-object-factory.ts
@@ -161,7 +161,7 @@ export class SchemaObjectFactory {
 
       const schemaCombinators = ['oneOf', 'anyOf', 'allOf'];
       if (schemaCombinators.some((key) => key in property)) {
-        delete (property as SchemaObjectMetadata).type;
+        (property as SchemaObjectMetadata).type = 'object';
       }
       return property as ParameterObject;
     });
--- a/test/services/schema-object-factory.spec.ts
+++ b/test/services/schema-object-factory.spec.ts
@@ -136,6 +136,7 @@ describe('SchemaObjectFactory', () => {
           profile: {
             description: 'Profile',
             nullable: true,
+            type: 'object',
             allOf: [
               {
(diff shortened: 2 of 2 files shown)


lib/services/schema-object-factory.ts
```diff
@@ -161,7 +161,7 @@ export class SchemaObjectFactory {
 
       const schemaCombinators = ['oneOf', 'anyOf', 'allOf'];
       if (schemaCombinators.some((key) => key in property)) {
-        delete (property as SchemaObjectMetadata).type;
+        (property as SchemaObjectMetadata).type = 'object'
```
gausnes, line 164: Didn't label as a breaking change as this is adding not removing properties. Based on my testing OpenAPI validators were happy with this addition.

TiBarification:
Is it possible to add these changes to main repo?

fredx21:
We are also impacted by this problem. What is required for the merge?

PR #1886, opened 2022-04-11
feat(): routing platform independence

flamewow:
## PR Checklist

## PR Type

Makes nestjs/swagger module routing platform independent (i.e. you don't need to install swagger-ui-express, or fastify-swagger anymore)
+ SwaggerCustomOptions are identical for both routing platforms
+ as bonus feature - add '-yaml' to swagger baseUrl to get yaml version of swagger document
+ added code snippet for manual e2e test of nestjs/swagger 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] Yes
- [ ] No


jmcdo29:
I think this approach is really cool. Does it have complete feature parity with swagger-ui-express and fastify-swagger? And this would require us to start maintaining our own swagger UI code, correct?

flamewow:
> I think this approach is really cool. Does it have complete feature parity with swagger-ui-express and fastify-swagger? And this would require us to start maintaining our own swagger UI code, correct?

It keeps all features of swagger-ui-express only (https://github.com/nestjs/swagger/pull/1886/files#diff-0b735d6eab8fb99376e025f69e9bbc23de289ee960da4d11e3e7b7965f9fccf3R1), FastifySwaggerCustomOptions are deprecated.

We'll need to maintain the layer between [swagger-ui](https://github.com/swagger-api/swagger-ui) and nestjs/swagger.
According to [swagger-ui-express](https://github.com/scottie1984/swagger-ui-express/commits/master) it's not updated very often, but yeah - we'll need to check for swagger-ui-express updates from time to time.

e2e/express.e2e-spec.ts
```diff
@@ -0,0 +1,70 @@
...
+      let api = await SwaggerParser.validate(document as any);
+      console.log(
+        'API name: %s, Version: %s',
+        api.info.title,
+        api.info.version
+      );
+      expect(api.info.title).toEqual('Cats example');
+    } catch (err) {
+      console.log(doc);
+      expect(err).toBeUndefined();
```
thiagomini, line 51: I Suppose this was a trick to have the `doc` logged only in the cases where the test fails? It took me a while to get this (if I understood this correctly), so I would suggest to make this test a bit easier to understand by removing the try..catch block and always logging the doc. Something like:
```ts
 const document = SwaggerModule.createDocument(app, builder.build());
 const doc = JSON.stringify(document, null, 2);
 
 console.log(doc)
 
 const api = await SwaggetParser.validate(document)
 console.log('API name: ...')
 expect(api.info.title).toEqual('Cats example');
```  

Btw, I would also suggest to have the 'Cats example' string inside a constant to avoid matching against magical strings. you could replace it by `TEST_API_TITLE`

e2e/manual-e2e.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const port = 4001;
+const host = 'localhost';
+const docRelPath = '/api-docs';
+
+async function bootstrap() {
+  const app = await NestFactory.create<NestFastifyApplication>(
+    ApplicationModule,
+    new FastifyAdapter()
+    // new ExpressAdapter()
```
thiagomini, line 20: reminder to remove this comment

flamewow: manual-e2e has a capability to quickly switch between Fastify & Express by commenting/uncommenting Adapter

that was the initial idea
Do u think it's not ok?

e2e/express.e2e-spec.ts
flamewow: it's almost a blind copy of existing fastify.e2e-spec.ts =)

lib/swagger-ui/swagger-ui.ts
```diff
@@ -0,0 +1,72 @@
+import * as swaggerUi from 'swagger-ui-dist';
+import { favIconHtml, htmlTemplateString, jsTemplateString } from './constants';
+import { SwaggerCustomOptions } from '../interfaces';
```
kamilmysliwiec, line 3: This file could use some refactoring:

- consts to functions
- add missing TS types
- make variables more descriptive/add comments in "buildJSInitOptions"

lib/swagger-module.ts
```diff
@@ -35,87 +40,87 @@ export class SwaggerModule {
...
-    );
+
     if (httpAdapter && httpAdapter.getType() === 'fastify') {
-      return this.setupFastify(
-        finalPath,
-        httpAdapter,
-        document,
-        options as FastifySwaggerCustomOptions
-      );
+      const fastifyApp = app as NestFastifyApplication; // TODO: figure out cleaner solution
```
kamilmysliwiec, line 47: that's fine 👍

lib/interfaces/swagger-custom-options.interface.ts
```diff
@@ -16,45 +12,3 @@ export interface ExpressSwaggerCustomOptions
   url?: string;
   urls?: Record<'url' | 'name', string>[];
 }
-
-export interface FastifySwaggerCustomOptions
```
kamilmysliwiec, line 20: Could we add an extra, temporary compatibility layer that would allow us to merge this PR without publishing a new major release? A layer that would still let library consumers pass `FastifySwaggerCustomOptions` etc and would convert them (options) to whatever interface we require them to pass now. We could remove this layer later on in the next release. Let me know if it's feasible 🙌  @flamewow

flamewow: we can have a compatibility layer for FastifySwaggerCustomOptions.uiConfig

but for  
` staticCSP?: boolean | string | Record<string, string | string[]>;
  transformStaticCSP?: (header: string) => string;
  uiHooks?: {
    onRequest?: Function;
    preHandler?: Function;
  };`
  there's no equivalent functionality in swagger-ui-express (don't know if it's even needed tbh)

Do u think adding partial compatibility layer, i.e. old fastify nestjs/swagger use cases won't break and those features (as uiHooks) just won't work is ok?
Or it worth making it a major upgrade forcing everyone to consciously migrate their swagger initialization?

e2e/manual-e2e.ts
thiagomini: I think it's only a nit to be honest, but I personally dislike commented code if there's an alternative, for instance, using constants to control this flow. For instance:
```ts
   const EXP = 'express'
   const FST = 'fastify'
   const getAdapter = {
      express: () => new ExpressAdapter,
      fastify: () => new FastifyAdapter
    }


const app = await NestFactory.create(ApplicationModule, getAdapter[EXP]())   
  
```

e2e/express.e2e-spec.ts
thiagomini: Hmmm I see, but if the `validate` function throws an error the test would fail as expected, right?

lib/interfaces/swagger-custom-options.interface.ts
kamilmysliwiec: @flamewow I've seen codebases where `uiHooks` were used for Basic authentication (to access swagger docs). Maybe we can somehow pass them down? Could you take a look at `fastify-swagger` and try to mimic some of its functionality (totally understandable if having 100% parity isn't doable)?

package.json
```diff
@@ -41,7 +47,7 @@
     "eslint": "8.12.0",
     "eslint-config-prettier": "8.5.0",
     "eslint-plugin-import": "2.26.0",
-    "express": "4.17.3",
+    "express": "^4.17.3",
```
micalevisk, line 50:
```suggestion
    "express": "4.17.3",
```

package.json
```diff
@@ -17,12 +17,18 @@
...
+    "start:dev": "nest start --watch",
+    "start:debug": "nest start --watch --debug"
   },
   "dependencies": {
     "@nestjs/mapped-types": "1.0.1",
+    "js-yaml": "^4.1.0",
     "lodash": "4.17.21",
-    "path-to-regexp": "3.2.0"
+    "path-to-regexp": "3.2.0",
+    "swagger-ui-dist": "^4.10.3"
```
micalevisk, line 31:
```suggestion
    "swagger-ui-dist": "4.10.3"
```

package.json
```diff
@@ -17,12 +17,18 @@
...
-    "release": "release-it"
+    "release": "release-it",
+    "---execution---": "",
+    "start": "nest start",
+    "start:dev": "nest start --watch",
+    "start:debug": "nest start --watch --debug"
   },
   "dependencies": {
     "@nestjs/mapped-types": "1.0.1",
+    "js-yaml": "^4.1.0",
```
micalevisk, line 28: let's pin this as dependencies upgrades are handled by Renovatebot

```suggestion
    "js-yaml": "4.1.0",
```

lib/swagger-module.ts
```diff
@@ -35,87 +40,87 @@ export class SwaggerModule {
     };
   }
 
-  public static setup(
-    path: string,
-    app: INestApplication,
-    document: OpenAPIObject,
-    options?: SwaggerCustomOptions
-  ) {
+  private static serveStatic(finalPath: string, app: INestApplication) {
```
delucca, line 43:
```suggestion
  private static serveStatic(finalPath: string, app: INestApplication | NestExpressApplication | NestFastifyApplication) {
```
I think you can use these types for the `app` param to avoid the `as NestFastifyApplication` below

lib/swagger-module.ts
delucca: I've added a suggestion in the function signature that maybe is a way to remove the need of the `as` here :)

lib/swagger-ui/swagger-ui.ts
```diff
@@ -0,0 +1,72 @@
...
+  );
+
+  json = json.replace(new RegExp('"' + placeholder + '"', 'g'), function (_) {
+    return fns.shift();
+  });
+
+  return `let options = ${json};`;
+};
+
+export const buildSwaggerHTML = function (
```
delucca, line 27: This is basically almost the same as:
```suggestion
export function buildSwaggerHTML(
```

lib/swagger-ui/swagger-ui.ts
```diff
@@ -0,0 +1,72 @@
...
+  baseUrl,
+  swaggerDoc,
+  options: SwaggerCustomOptions
+) {
+  const customCss = options?.customCss ?? '';
+  const customJs = options?.customJs ?? '';
+  const customfavIcon = options?.customfavIcon ?? false;
+  const customSiteTitle = options?.customSiteTitle ?? 'Swagger UI';
+  const customCssUrl = options?.customCssUrl ?? '';
```
delucca, line 36: We can simplify thise whole block to:
```suggestion
  const { customCss, customJs, customfavIcon, customSiteTitle, customCssUrl } = options;
```

And them assign the defaults where you're using them, with the `??` operator

delucca (commented):
I've added some minor comments 😄

Great job 🚀

package.json
```diff
@@ -38,6 +38,7 @@
     "@nestjs/platform-express": "8.4.4",
     "@nestjs/platform-fastify": "8.4.4",
     "@types/jest": "27.4.1",
+    "@types/js-yaml": "^4.0.5",
```
micalevisk, line 41:
```suggestion
    "@types/js-yaml": "4.0.5",
```

lib/swagger-module.ts
flamewow: your suggestion doesn't work

I've came up with a bit cleaner solution than before

lib/interfaces/swagger-custom-options.interface.ts
flamewow: @kamilmysliwiec tried to make uiHooks work with the new approach, unfortunately no success here.
In order to keep feature compatibility (uiHooks specifically) I'll need to copy good chunk of fastify-swagger code (fastify routing + plugin registration etc) which makes the whole point of the PR questionable.

Added backward compatibility layer for uiConfig.

kamilmysliwiec: > I'll need to copy good chunk of fastify-swagger code (fastify routing + plugin registration etc)

I think that's _fine_(?). How much code would we have to move? Just double-checked and it seems that the only thing we'd need is this https://github.com/fastify/fastify-swagger/blob/master/lib/routes.js#L89 (adding `hooks` to the route _if registered_), but correct me if I'm wrong .

Perhaps we could just add another condition in the `serveDocuments` and if platform === fastify then grab Fastify adapter and use `.route()` to register hooks? Obviously it would make this method less "platform-agnostic" (currently it only uses `HttpAdapter` which is great) but at least would let folks keep using `uiHooks` which seems to be crucial

flamewow: @kamilmysliwiec updated (included serveDocumentsFastify function to handle documents in fastify case, supports uiHooks)
Added sample code to showcase uiHooks and ability to set custom CSS (in fastify case).

P.S. it is not as bad as I anticipated, guess I was too focused on making it truly universal )

lib/swagger-ui/swagger-ui.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const jsInitOptions = buildJSInitOptions(swaggerInitOptions);
+  return jsTemplateString.replace('<% swaggerOptions %>', jsInitOptions);
+}
+
+/**
+ * Stores absolute path to swagger-ui assets
+ */
+export const swaggerAssetsAbsoluteFSPath = swaggerUi.getAbsoluteFSPath();
+console.log(swaggerAssetsAbsoluteFSPath);
```
kamilmysliwiec, line 28:
```suggestion
console.log(swaggerAssetsAbsoluteFSPath);
```
Is this intentional? @flamewow

package.json
```diff
@@ -17,12 +17,18 @@
     "test:dev": "jest --watch",
     "test:e2e": "jest --config e2e/jest-e2e.json",
     "prerelease": "npm run build",
-    "release": "release-it"
+    "release": "release-it",
+    "---execution---": "",
```
kamilmysliwiec, line 21:
```suggestion
    "---execution---": "",
```
should we rename it to manual-test?

lib/interfaces/swagger-custom-options-legacy.instace.ts
```diff
@@ -0,0 +1,67 @@
+interface _CommonSwaggerCustomOptions {
```
kamilmysliwiec, line 1: Can we rename this file to `legacy-custom-options.interfaces.ts`?

lib/interfaces/swagger-custom-options-legacy.instace.ts
```diff
@@ -0,0 +1,67 @@
...
+  useGlobalPrefix?: boolean;
+}
+
+export interface ExpressSwaggerCustomOptions
```
kamilmysliwiec, line 5: Should this be marked as `@deprecated` as well?

lib/swagger-module.ts
```diff
@@ -35,87 +45,107 @@ export class SwaggerModule {
...
 
-  private static setupExpress(
-    path: string,
-    app: INestApplication,
-    document: OpenAPIObject,
-    options?: ExpressSwaggerCustomOptions
-  ) {
+    // START: fastify backward compatibility layer
+    const { customOptions, extra } = processSwaggerOptions(options);
+    // END: fastify backward compatibility layer
```
kamilmysliwiec, line 115: Just to make sure that I'm following - once we decide to remove the backward compatibility layer, we can simply remove these lines? And then `options` === `customOptions`? What about `extra`?

lib/swagger-module.ts
```diff
@@ -35,87 +45,107 @@ export class SwaggerModule {
...
-      require('swagger-ui-express')
-    );
-    const swaggerHtml = swaggerUi.generateHTML(document, options);
-    app.use(path, swaggerUi.serveFiles(document, options));
 
-    httpAdapter.get(path, (req, res) => res.send(swaggerHtml));
-    httpAdapter.get(path + '-json', (req, res) => res.json(document));
-  }
+    // START: fastify backward compatibility layer
+    const IS_FASTIFY = httpAdapter && httpAdapter.getType() === 'fastify';
```
kamilmysliwiec, line 125:
```suggestion
    const isFastify = httpAdapter && httpAdapter.getType() === 'fastify';
```
we use CONSTANT_CASE (`IS_FASTIFY`) for constants

lib/backward-compatilibity-layer.ts
```diff
@@ -0,0 +1,127 @@
+import {
```
kamilmysliwiec, line 1: The philosophy behind introducing the backward compatibility layer was to avoid introducing breaking changes (in options interfaces). In the next major release, we'll simply remove the compatibility layer BUT features like `uiHooks` should IMO still keep working as expected @flamewow (otherwise we won't be able to guarantee 100% feature parity)

flamewow: hooks are the feature of fastify, i.e. developers won't be able to simply swap fastify to express adapter in a NestFactory.create method params and expect it to work the same way.

Initial idea was to achieve behavior like "upon swapping adapter dev should be able to get the same behavior from nestjs/swagger as it was with a different adapter without any modifications to swagger configuration" in a long run.
Otherwise, it's not truly platform-independent, the only thing we achieved is we got rid of extra dependency (express-swagger-ui, fastify-swagger).

Do you want to keep that uiHooks as an exclusive fastify-only feature in a long run, or to come up with a universal solution that would work for express & fastify alike?

lib/swagger-ui/swagger-ui.ts
flamewow: nope, thanks for spotting it, forgot to remove it after debugging

package.json
flamewow: do you mean the section ("---execution---") or section & scripts (start -> 'manual-test')?

kamilmysliwiec: @flamewow updated code suggestion

lib/backward-compatilibity-layer.ts
kamilmysliwiec: @flamewow can you run some tests with adding basic auth in front of Swagger UI (for fastify) using a simple middleware? Perhaps with this "universal" approach `uiHooks` can be swapped with something else (middleware-based approach)

lib/swagger-module.ts
flamewow: extra is the diff that is currently supported only in fastify (uiHooks).
Let's clarify the end goal (thread under [lib/backward-compatilibity-layer.ts](https://github.com/nestjs/swagger/pull/1886/files/92e85c9c2d2c0f979cd9d4244d492bdaae706b6f#diff-45b88e46065760c5d355b77cd01bb5155c53012ffcc7a5131275639099de34fe))
and I would be able to answer this

flamewow: 1. set setup method param 'options' type to SwaggerCustomOptions
2. drop usage of processSwaggerOptions (use customOptions = options)
3. drop usage of serveDocumentsFastify
4. remove backward-compatility-layer.ts file

and we need to gradually update documentation on https://docs.nestjs.com/openapi/introduction

flamewow:
added minor fixes for fastify, the issue appeared when '/' is appended to swagger URL
i.e. `/api-docs` -> `/api-docs/`
fastify was not serving html in `/api-docs/` case (while express works fine with it)
to keep identical behavior added the fix

package.json
```diff
@@ -42,30 +49,25 @@
     "eslint-config-prettier": "8.5.0",
     "eslint-plugin-import": "2.26.0",
     "express": "4.17.3",
-    "fastify-swagger": "5.1.1",
+    "fastify-static": "4.6.1",
```
0xslipk, line 52: Could you please use `@fastify/static@5.0.0` instead of?. fastify-static has been deprecated. fastify/fastify#3733

rumourscape:
When will this be merged so that I can use Swagger?

alexNerazim:
any update?  still working on that? Thanks!

PR #1915, opened 2022-05-01
refactor(types): improved security scheme

xTCry:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


e2e/fastify.e2e-spec.ts
```diff
@@ -27,8 +27,8 @@ describe('Fastify Swagger', () => {
       .addBearerAuth()
       .addOAuth2()
       .addApiKey()
-      .addApiKey({ type: 'apiKey' }, 'key1')
-      .addApiKey({ type: 'apiKey' }, 'key2')
```
jmcdo29, line 31: This would make it a breaking change, but would it make more sense to bring the name to be the first parameter instead of the options?

lib/document-builder.ts
```diff
@@ -136,34 +137,34 @@ export class DocumentBuilder {
     return this;
   }
 
-  public addApiKey(
-    options: Optional<SecuritySchemeObject, 'type'> = {},
-    name = 'api_key'
+  public addBasicAuth(
+    options: Partial<Omit<HttpSchemeObject, 'type'>> = {},
+    name = 'basic'
```
jmcdo29, line 142: Making the change I mentioned above here so it would be
```suggestion
		name = 'basic',
		options: Partial<Omit<HttpSchemeObject, 'type'>> = {}
```

xTCry:
@jmcdo29 Yes, I agree. But this is likely to be a breaking change

micalevisk:
@xTCry I guess we could keep both and mark the old one as deprecated

kamilmysliwiec:
What's the reason for this change?

e2e/fastify.e2e-spec.ts
xTCry: @jmcdo29 , yep, you're right. Fixed

lib/document-builder.ts
xTCry: @jmcdo29, Yes, this can also be added, but is it better in a minor update?..

xTCry:
I apologize for the long answer.

> What's the reason for this change?
@kamilmysliwiec

This is a typing improvement, based on the official Swagger documentation. It is also simplified to access some methods without specifying unnecessary parameters.

For example, why write `.addApiKey({ type: 'apiKey'}, 'key1')`, if the name of the method speaks for itself that it accepts only the type `apiKey`.
Also, `HttpSchemaObject` accepts `bearerFormat`, `OpenIdConnectSchemeObject` can use `openIdConnectUrl`. But they are not used in other ways

car1ot:
Accept this pull request

PR #2023, opened 2022-08-10
perf(swagger-module): lazy load swagger-ui-dist module

kaioduarte:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- The `swagger-ui-dist` module is loaded before its utilization, and it loads two large files (listed below), causing some overhead on the app initialization (~800ms in a medium app).
  - https://github.com/swagger-api/swagger-ui/blob/master/swagger-ui-dist-package/index.js#L2-L3, 
  -  https://github.com/swagger-api/swagger-ui/blob/master/dist/swagger-ui-standalone-preset.js
  -  https://github.com/swagger-api/swagger-ui/blob/master/dist/swagger-ui-bundle.js

Issue Number: N/A

## What is the new behavior?
- The `swagger-ui-dist` module is loaded only when serving the Swagger HTML (`SwaggerModule.serveStatic`).

- [ ] Yes
- [x] No

## Other information

Bubbleprof from Clinicjs using the E2E app.

| Before  | After  | 
|---|---|
| image  | image |

### DIFF
--- a/lib/swagger-ui/swagger-ui.ts
+++ b/lib/swagger-ui/swagger-ui.ts
@@ -1,4 +1,3 @@
-import * as swaggerUi from 'swagger-ui-dist';
 import { favIconHtml, htmlTemplateString, jsTemplateString } from './constants';
 import { OpenAPIObject, SwaggerCustomOptions } from '../interfaces';
 import { buildJSInitOptions } from './helpers';
@@ -21,10 +20,21 @@ export function buildSwaggerInitJS(
   return jsTemplateString.replace('<% swaggerOptions %>', jsInitOptions);
 }
 
+let swaggerAssetsAbsoluteFSPath;
+
 /**
- * Stores absolute path to swagger-ui assets
+ * Returns the absolute path to swagger-ui assets.
  */
-export const swaggerAssetsAbsoluteFSPath = swaggerUi.getAbsoluteFSPath();
+export function getSwaggerAssetsAbsoluteFSPath() {
+  if (!swaggerAssetsAbsoluteFSPath) {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const swaggerUi = require('swagger-ui-dist');
+    swaggerAssetsAbsoluteFSPath = swaggerUi.getAbsoluteFSPath();
+  }
+
+  return swaggerAssetsAbsoluteFSPath;
+}
+
 /**
  * Used to build swagger-ui custom html
  */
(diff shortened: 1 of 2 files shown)


lib/swagger-ui/swagger-ui.ts
```diff
@@ -21,10 +20,21 @@ export function buildSwaggerInitJS(
   return jsTemplateString.replace('<% swaggerOptions %>', jsInitOptions);
 }
 
+let swaggerAssetsAbsoluteFSPath;
```
micalevisk, line 23:
```suggestion
let swaggerAssetsAbsoluteFSPath: string | undefined;
```

kaioduarte: Done, thanks!

micalevisk (approved):
this will be great for serverless apps! thanks

kamilmysliwiec:
Could we use `await import()` instead of require 🤔 ?

kaioduarte:
@kamilmysliwiec we could, but that will require users to add `await` to the `SwaggerModule.setup` method.

kamilmysliwiec:
So this is something that makes me a little concerned. If we wanted to support ESM (eventually), we'd have to use async import anyway (for lazy imports)

kaioduarte:
I'm fine doing that change. However, do I need to update any doc to warn users about the need to await the setup method?

micalevisk:
wouldn't that be a breaking change? Maybe we could keep `require()` + `// TODO(v7): change to import statement.`

kamilmysliwiec:
Yeah that would be a breaking change, and that's why I'm concerned too. If we stick with the current implementation (top-level require), even if we decide to move to ESM one day, we wouldn't have to introduce any breaking changes

micalevisk:
yeah, indeed.

Another solution that I can think of:

```js
SwaggerModule.setup('api', app, document) // this won't use lazy loading

await SwaggerModule.setup('api', app, document, { lazy: true })
```

then in th next major release of `@nestjs/swagger` we'll remove that parameter. Not sure if this is feasible tho

kaioduarte:
Hey @kamilmysliwiec, what approach should we take? I'm not an ESM expert, but I'm happy to help and get this PR merged 🙂

kaioduarte:
🏓

kamilmysliwiec:
Let's ignore ESM for now. LGTM

PR #2068, opened 2022-09-01
feature(document-builder): improved types of security scheme methods

xTCry:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

This is a typing improvement, based on the official Swagger documentation. It is also simplified to access some methods without specifying unnecessary parameters.
For example, why write `.addApiKey({ type: 'apiKey'}, 'key1')`, if the name of the method speaks for itself that it accepts only the type apiKey.
Also, `HttpSchemaObject` accepts `bearerFormat`, `OpenIdConnectSchemeObject` can use `openIdConnectUrl`. But they are not used in other ways

Updated the order of arguments in methods. (Added deprecation for old)
https://github.com/nestjs/swagger/pull/1915#discussion_r862519173

Issue Number: N/A

- [ ] Yes
- [x] No


xTCry:
@micalevisk, can you please review this PR

xTCry:
@kamilmysliwiec can you accept this PR?

lib/document-builder.ts
```diff
@@ -110,74 +115,147 @@ export class DocumentBuilder {
...
   ): this {
-    this.addSecurity(name, {
+    if (typeof nameOrOptions === 'object') {
+      [nameOrOptions, optionsOrName] = [optionsOrName, nameOrOptions];
+    }
+    this.addSecurity(nameOrOptions as string, {
       type: 'oauth2',
       flows: {},
-      ...options
+      ...(optionsOrName as any)
```
micalevisk, line 169: I guess we need to change the signature of `addSecurity` instead of using type assertion here and in other places, right?

xTCry: @micalevisk, sorry, I still don't understand what you meant. 😨

micalevisk: I mean, can't we avoid the `as any` here?

xTCry: @micalevisk I'm sorry, but I really don't have time to improve it. I have already tried to update the typing so that it is more correct for developers outside the module.
And this part that you are asking to fix will not be so important outside, but only inside this module. This can be marked as TODO at least and make these changes later.

micalevisk: it's fine

```suggestion
      // TODO: avoid using type assertion to 'any' below
      ...(optionsOrName as any)
```

xTCry:
@kamilmysliwiec Can this improvement see the light in the next release?

rtkid-nt:
Good PR

xTCry:
@micalevisk, types updated

PR #2121, opened 2022-10-14
feat: add `.vo.ts` to `dtoFileNameSuffix`

ZanMinKian:
Thanks to `ValidatetionPipe`, we can use `dto` (Data Transfer Object) to validate the __request__ params. Similar with `dto`, `vo` (View Object) should be used to serialize __reponse__ body (powered by `ClassSerializerInterceptor `).

`student.dto.ts`
```ts
export class StudentDto {
  @IsString()
  name!: string;
  @IsString()
  password!: string;
}
```
`student.vo.ts`
```ts
@Exclude()
export class StudentVo {
  @Expose()
  name!: string;
  // @Expose() // Prevent developer leaking password by mistake. So should not add @Expose()
  password!: string;
}
```
`student.controller.ts`
```ts
@Controller('student')
export class StudentController {
  @Post('update')
  update(student: StudentDto): Promise<StudentVo> {
    // ...
  }
}
```

This PR encourage developer to  separate __request__ params objects and __response__ body objects and put them into their own file. Swagger api doc will make  frontend developers more clear about what request and response look like.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/lib/plugin/merge-options.ts
+++ b/lib/plugin/merge-options.ts
@@ -10,7 +10,7 @@ export interface PluginOptions {
 }
 
 const defaultOptions: PluginOptions = {
-  dtoFileNameSuffix: ['.dto.ts', '.entity.ts'],
+  dtoFileNameSuffix: ['.dto.ts', '.vo.ts', '.entity.ts'],
   controllerFileNameSuffix: ['.controller.ts'],
   classValidatorShim: true,
   dtoKeyOfComment: 'description',


micalevisk:
that configuration is pretty easy to change, and it's well documented. Thus, I don't see any reason to change the current behavior. Also, as the default, I think that the less files the better.

ZanMinKian:
In most of enterprise projects, request params type and response body type are the most important  things for a HTTP api.
Nest has a `dto` concept (inspired by Java?) for validating request params. But what's for serializing response body? I don't see any documents explain the best practice to handle response body properly.

I can put all the `dto`s into `.dto.ts` files. `@nestjs/swagger` will generate api doc containing request params types.  What about the response body types? Should I put them into `.dto.ts` files too? Seems not good. So I use the `vo` concept (inspired by Java) to tell developer: "Hey, you should put your response types into `.vo.ts` files. I, `@nestjs/swagger`, can handle it properly".

Moreover, I want to create a PR to `@nestjs/cli` to make `nest` command line support `--plugins` option. We can run `nest build --plugins=@nestjs/swagger` without `nest-cli.json` file. Zero-configs is better. Zero-configs require `@nestjs/swagger` provide the most convenient configs by default.

kamilmysliwiec:
I agree with @micalevisk. If you need to support more suffixes, just add them to the config. That's why we exposed this configuration.

PR #2257, opened 2023-01-27
fix: add getSchemaPath into shim

Inoir:
## PR Type

- [x ] Bugfix

getSchemaPath is missing in shims.

## What is the new behavior?
If you use decorators and also getSchemaPath in ui and use shim, doesnt find the function anymore

- [ ] Yes
- [ x] No

Only for shim. nothing changed

### DIFF
--- a/lib/extra/swagger-shim.ts
+++ b/lib/extra/swagger-shim.ts
@@ -165,5 +165,5 @@ export function PickType() {
   return () => {};
 }
 export function getSchemaPath() {
-  return () => {}
+  return () => ""
 }
--- a/lib/extra/swagger-shim.ts
+++ b/lib/extra/swagger-shim.ts
@@ -164,3 +164,6 @@ export function PartialType() {
 export function PickType() {
   return () => {};
 }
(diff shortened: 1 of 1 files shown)


lib/extra/swagger-shim.ts
```diff
@@ -164,3 +164,6 @@ export function PartialType() {
 export function PickType() {
   return () => {};
 }
+export function getSchemaPath() {
+  return () => {}
```
micalevisk, line 168: `getSchemaPath` returns a string, so I guess this should be
```suggestion
  return ''
```

https://github.com/nestjs/swagger/blob/ae525047c27c4c0c4383b4db8ab6f433a510be4b/lib/utils/get-schema-path.util.ts#L3-L6

Inoir: oh yes this was brain death bad copy paste

micalevisk (approved):

Tony133 (approved):

kamilmysliwiec:
lgtm

PR #2351, opened 2023-03-21
Refactoring: koa adapter bug

jfkz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

There is an issue in nest-koa-adapter which caused by non-standart calls inside @nestjs/swagger.

I rewrited them to abstract class calls so we can implement other adapters as well.

Issue Number: https://github.com/seidelmartin/nest-koa-adapter/issues/23

## What is the new behavior?

Instead of using non-standart (framework-depend) assigning I rewrite code to use abstract class implementations.

- [ ] Yes
- [x] No


jmcdo29:
Rather than adding the koa support directly to `@nestjs/swagger`, as we don't have an "official" koa adapter, could we make these methods `protected` so a community swagger adapter for koa could be written and used in junction with Nest's `@nestjs/swagger`? Essentially, still use all of Nest's decorators, but make use of a `KoaSwaggerModule` provided by the community

jfkz:
I understand. There is a problem with Koa cause their interface contains `type` and `response` as a fields by their architecture.

My goal was to shift from non-standard framework specific implementation to more abstract solution which can be used in any future connectors.

jmcdo29:
> My goal was to shift from non-standard framework specific implementation to more abstract solution which can be used in any future connectors.

I think the changes to using `httpAdapter` instead of `res` directly are nice. It was just the addition of the `adapter.type === 'koa'` I was calling out with my last comment

PR #2371, opened 2023-04-03
fix: allow missing paramtypes metadata when extracting method parameter types

effervescentia:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using `@nestjs/swagger` in environments that do not support TypeScript's `emitDecoratorMetadata` configuration this library will throw an exception if any parameter decorator is discovered because the `design:paramtypes` metadata array it expects to find in `parameter-metadata-accessor.ts` will not exist.

Issue Number: N/A

## What is the new behavior?

With this change the `types` array – which is searched for the default type of each parameter – will be an empty array if `design:paramtypes` does not exist on the controller method. This is a safe value to default to as it will at worst result in routes with no schema details for their parameters.

- [ ] Yes
- [x] No

### DIFF
--- a/lib/services/parameter-metadata-accessor.ts
+++ b/lib/services/parameter-metadata-accessor.ts
@@ -38,7 +38,7 @@ export class ParameterMetadataAccessor {
       PARAMTYPES_METADATA,
       instance,
       method.name
-    );
+    ) ?? [];
     const routeArgsMetadata: ParamsMetadata =
       Reflect.getMetadata(
         ROUTE_ARGS_METADATA,
--- a/test/services/parameter-metadata-accessor.spec.ts
+++ b/test/services/parameter-metadata-accessor.spec.ts
@@ -1,0 +1,40 @@
+import { Controller, Get, Query } from '@nestjs/common';
+import { PARAMTYPES_METADATA } from '@nestjs/common/constants';
+import { ParameterMetadataAccessor } from '../../lib/services/parameter-metadata-accessor';
+
+describe('ParameterMetadataAccessor', () => {
+  @Controller()
+  class UserController {
+    @Get()
+    findUser(@Query('userID') userID: string) { }
+  }
+
+  let parameterMetadataAccessor: ParameterMetadataAccessor;
+
+  beforeEach(() => {
+    parameterMetadataAccessor = new ParameterMetadataAccessor();
+  });
+
+  describe('explore', () => {
+    it('should handle missing paramtypes metadata', () => {
+      const controller = new UserController();
+      jest.spyOn(Reflect, 'getMetadata')
+        .mockReturnValueOnce(undefined)
+        .mockReturnValueOnce({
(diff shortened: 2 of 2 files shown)


kamilmysliwiec:
Just out of curiosity, why would anyone want to use this package without having the `emitDecoratorMetadata` enabled 🤔

effervescentia:
> Just out of curiosity, why would anyone want to use this package without having the `emitDecoratorMetadata` enabled 🤔

@kamilmysliwiec I didn't want to get too into my use-case too much to avoid confusing the bugfix 😛 

I'm just casually playing god by trying to get NestJS to work with ESM modules, `typeorm` and `swagger`
We're using your microservices framework (which was really nice to work with, though I think the documentation is lacking some key examples / details) to create a module + custom strategy + client that can be used to start and interface with a [logux](https://logux.io/) server. We'll probably make a blog post about it at some point in the future / open source, there's a lot of finicky details we needed to figure out to get everything working nicely.

Surprisingly enough there are only a couple of big hurdles to make the stack itself work nicely:
1. TypeScript + ESM does not work with most non-ESM dev tooling, and even some ESM-specific tooling (`ts-node`, `tsm`)
  - we're solving this by using [tsx](https://www.npmjs.com/package/tsx), which has only one drawback, **no support for emitting decorator metadata**
2. Explicitly adding `@Inject()` to injected constructor parameters
  - this is a bit annoying, but we are already doing it because they weren't being picked up by `jest` in some circumstances I believe

We already have a fairly complicated multi-layer monorepo where all `typeorm`-related models are in a separate package. This means they have been compiled already and correctly have all relevant metadata baked in already. I'm not sure if it becomes less feasible to integrate `typeorm` effectively without this.

As for the limitations of `tsx`, this is only imposed at development time. We generate our API schemas as part of our CI process, so not having the correct details at this point is inconsequential to our workflow. However we also use the `@Api*` decorators from `@nestjs/swagger` to describe every endpoint, parameter, query, etc so we actually end up having accurate details either way.

Hope this helps! I'd be happy to dive into any other details if it would be helpful.

Also check out this ESM-compatible NestJS boilerplate I setup using `tsx`. Since this doesn't have swagger setup the only important changes are the explicit `@Import()` decorators.
https://stackblitz.com/edit/node-nxsjdb?file=src/app.controller.ts

effervescentia:
hmm okay so with some more testing I discovered that despite the above change the `@Body()` decorator still causes some exception to be thrown and the root cause is not having the name of the parameter, which isn't available at runtime without the `emitDecoratorMetadata` configuration enabled... I didn't delve too deep down that rabbit hole because the error was less apparent than the one with the other parameter decorators.

However, we have already developed a library that allows us to generate OpenAPI schemas without running the NestJS application itself. As a result we don't actually need to have `@nestjs/swagger` setup in our `main.ts` file so I'll close this PR for now

I noticed you merged my other PR for the `@nestjs/microservices` package though, thanks for that! 😄

Issue #2481, opened 2023-06-16
Swagger UI options not working after upgrading to v7


BC-M:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

It seems that after the upgrade to v7.0.0 the options no longer work. I have checked the [docs for swagger-ui v5 ](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/) but I cannot see the differences in the options.

### Minimum reproduction code

-

### Steps to reproduce

```ts
    const document = SwaggerModule.createDocument(app, config);
    SwaggerModule.setup('docs', app, document, {
      swaggerOptions: {
        persistAuthorization: true,
        tagsSorter: 'alpha',
        operationsSorter: 'alpha',
        defaultModelsExpandDepth: -1,
        defaultModelExpandDepth: -1,
      },
    });
  }
```

### Expected behavior

It should be the same as in v6.

### Package version

7.0.0

### NestJS version

10.0.0

### Node.js version

18.14.2

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [X] Linux

kamilmysliwiec:
I temporarily reverted back to v4 (in `@nestjs/swagger@7.0.1`).

@flamewow would you take a look at what would it take to upgrade `swagger-ui-dist` to v5 with our custom templates? 👀

joshunrau:
I am having a similar issue with other settings in`SwaggerCustomOptions` as well. I am trying to add some custom static assets, but this does not seem to have any effect (I see in the browser they are not being fetched). 

This is the code in question. I tried updating to 7.0.1 but this did not fix the issue.

```typescript
  app.useStaticAssets(path.resolve(__dirname, '..', 'public'));

  const config = new DocumentBuilder()
    .setTitle('The Douglas Data Bank')
    .setDescription('Documentation for the Douglas Data Bank API')
    .setLicense('AGPL-3.0', 'https://www.gnu.org/licenses/agpl-3.0.txt')
    .setVersion('1.0')
    .addTag('Auth')
    .build();

  const document = SwaggerModule.createDocument(app, config);

  SwaggerModule.setup('/', app, document, {
    customCssUrl: '/swagger.css',
    customfavIcon: '/favicon.ico',
    customSiteTitle: 'The Douglas Data Bank API'
  });
```

joshunrau:
Okay I figured out the issue I will make a PR

kamilmysliwiec:
[...]

PR #2392, opened 2023-04-15
perf: added possibility to lazy initialize the document

H4ad:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

Currently, NestJS builds the Swagger document synchronously on startup, which it's a waste of time since we don't care about the swagger routes until we need to make a call to it.

## What is the new behavior?

So, I added an option to lazy initialize the Swagger document, the document will only be built when we have some request for it.

In my API, it reduced about 70~75ms from initialization, it will vary according to the API and the amount of routes but now it will only take 2ms to initialize the Swagger Module.

### Usage

Now, to use the new lazy loading feature, just change these lines:

```diff
const swaggerDocument = new DocumentBuilder()
  .setTitle(config.get<string>('SWAGGER_TITLE') ?? 'Base')
  .setDescription(config.get<string>('SWAGGER_DESCRIPTION') ?? 'The base API')
  .setVersion(config.get<string>('VERSION') ?? '1.0.0')
  .addTag('main')
  .addBearerAuth()
  .build();

-const document = SwaggerModule.createDocument(app, swaggerDocument);
+const document = () => SwaggerModule.createDocument(app, swaggerDocument);

SwaggerModule.setup('/swagger', app, document, {
  swaggerOptions: { docExpansion: 'none' },
});
```

Just one line to use it.

### Reproduction

To use it: https://stackblitz.com/edit/nestjs-starter-bql4af?file=src/main.ts

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #2487, opened 2023-06-17
fix: issue where custom assets are not injected in html

joshunrau:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Options passed to `SwaggerModule.setup` are ignored. The `customOptions` object passed to `buildSwaggerHTML` is malformatted. Custom parameters that should be contained within the base of `customOptions` are actually contained within `swaggerOptions`, which itself contains another object `swaggerOptions`. For example, this is the object passed to `buildSwaggerHTML` when you run `npm run start:dev`:

```js
{
  jsonDocumentUrl: '/api-docs-json',
  yamlDocumentUrl: '/api-docs-yaml',
  swaggerOptions: {
    customSiteTitle: 'Demo API - Swagger UI 1',
    swaggerOptions: {
      persistAuthorization: true,
      defaultModelsExpandDepth: -1,
      syntaxHighlight: [Object],
      tryItOutEnabled: true
    },
    customfavIcon: '/public/favicon.ico',
    customCssUrl: '/public/theme.css'
  }
}
```

I changed how the options are parsed from the `customOptions` object in `buildSwaggerHTML`.

Issue Number: #2481

## What is the new behavior?

I implemented a quick fix to resolve the bug. Now, extracted the options from a merger of `customOptions` and `customOptions.swaggerOptions`. This is not an ideal solution, but it seems this object structure is assumed in several other places in the code base, and applying a more systematic fix was non-trivial, as someone who is unfamiliar with the project.

- [ ] Yes
- [x] No

## Other information

I added tests for all of the custom assets so this will not inadvertently break again in the future.

### DIFF
--- a/e2e/express.e2e-spec.ts
+++ b/e2e/express.e2e-spec.ts
@@ -188,4 +188,69 @@ describe('Express Swagger', () => {
       expect(response.text.length).toBeGreaterThan(0);
     });
   });
+
+  describe('custom swagger options', () => {
+    const CUSTOM_CSS = 'body { background-color: hotpink !important }';
+    const CUSTOM_JS = '/foo.js';
+    const CUSTOM_JS_STR = 'console.log("foo")';
+    const CUSTOM_FAVICON = '/foo.ico';
+    const CUSTOM_SITE_TITLE = 'Foo';
+    const CUSTOM_CSS_URL = '/foo.css';
+
+    beforeEach(async () => {
+      const swaggerDocument = SwaggerModule.createDocument(
+        app,
+        builder.build()
+      );
+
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
Thanks for your contribution! I took a slightly different approach as it seems that the regression was introduced in this PR https://github.com/nestjs/swagger/pull/2392 where we pass the outer options object instead of the Swagger-specific custom options object down to the function you modified. Still, I'm gonna merge this PR as it adds some valuable e2e tests and will prevent us from introducing similar breaking changes in the future!

PR #2514, opened 2023-07-06
fix: support defining `properties` for query-arguments of type objects

Addono:
## PR Checklist

- [x] ~Docs have been added / updated (for bug fixes / features)~ _N/A_

## PR Type

- [x] Bugfix

When setting an object with a pre-defined list of properties as a Query-argument, the generated OpenAPI schema will have the `properties` key outside of the `schema` definition. This violates the OpenAPI spec.

https://github.com/addono/swagger/blob/038a366548beac5ffd392508e43225722d2184c6/e2e/api-spec.json#L362-L369

image

image

Issue Number: N/A

## What is the new behavior?

The `properties`-field is kept as a child of the `schema`-key.

https://github.com/Addono/swagger/blob/1e9a3386228e0bcb229027901e415b476dda7272/e2e/api-spec.json#L362-L373

image

image

- [x] No

## Other information

This PR contains two commits:
 - First commit adds an example to the E2E tests which show how the generated spec is failing. For this commit, the E2E tests fail. This commit is there to allow reproducing the old behaviour.
 - Second commit fixes the behaviour and updates the expected spec in the E2E tests accordingly.

When merging, use squash-merge to avoid having a commit in the history for which the tests are known to fail.

### DIFF
--- a/e2e/api-spec.json
+++ b/e2e/api-spec.json
@@ -359,6 +359,14 @@
             "name": "filter",
             "required": true,
             "in": "query",
+            "properties": {
+              "name": {
+                "type": "string"
+              },
+              "age": {
+                "type": "number"
+              }
+            },
             "schema": {
               "additionalProperties": true,
               "type": "object"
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
LGTM

PR #1949, opened 2022-06-21
feat(swagger): Provide URI version to operationIdFactory

MichaelMarner:
Adds a new optional parameter to operationIdFactory to allow providing the API version prefix added to the operation's path.

This allows the operationIdFactory to generate unique IDs in the case of the same controller method being used for multiple API versions. Default behavior is unchanged - the operation IDs will be of the form controllerKey_methodKey, but where necessary this can be customised using the URI version prefix.

The implementation here gets the versions that the method is for, and generates the URI prefixes. It then looks for the version prefixes in the paths returned from `RoutePathFactory`, and provides that information to operationIdFactory, so that unique IDs can be generated.

I'm not _super_ familiar with Nest internals, but this looks like the only real way to implement this without making more extensive changes to `nest-router`, which would have potentially wider side effects. 

Closes #1711

## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
When using Nest's API versioning feature, and having a controller method cover multiple versions, nestjs-swagger generates invalid OpenAPI spec, due to duplicate operation IDs.

It generates duplicate operation IDs because only the controller and method names are available to `operationIdFactory` - but in these cases the same controller and method is used multiple times.

Issue Number: #1711 

## What is the new behavior?

When URI versioning is used, an additional parameter containing the version is provided to `operationIdFactory`. This allows developers to provide unique operation IDs in the case of a method being used in multiple versions.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #2549, opened 2023-08-02
feat(swagger): Provide URI version to operationIdFactory 

deongroenewald:
This is based on the work completed in #1949. Those changes were reverted due to test failures.
This change adapts those changes to also support neutral versions which resolves the test failures.

Adds a new optional parameter to operationIdFactory to allow providing the API version prefix added to the operation's path.

This allows the operationIdFactory to generate unique IDs in the case of the same controller method being used for multiple API versions.

Default behavior is unchanged - the operation IDs will be of the form controllerKey_methodKey, but where necessary this can be customised using the URI version prefix.

Closes nestjs#2537

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

When using Nest's API versioning feature, and having a controller method cover multiple versions, `nestjs-swagger `generates invalid OpenAPI spec, due to duplicate operation IDs.

It generates duplicate operation IDs because only the controller and method names are available to `operationIdFactory - but in these cases the same controller and method is used multiple times.

Issue Number: 2537

## What is the new behavior?

When URI versioning is used, an additional parameter containing the version is provided to `operationIdFactory`. This allows developers to provide unique operation IDs in the case of a method being used in multiple versions.

- [ ] Yes
- [x] No

### DIFF
--- a/lib/interfaces/swagger-document-options.interface.ts
+++ b/lib/interfaces/swagger-document-options.interface.ts
@@ -1,7 +1,7 @@
 export type OperationIdFactory = (
   controllerKey: string,
   methodKey: string,
-  pathVersionKey?: string
+  version?: string
 ) => string;
 
 export interface SwaggerDocumentOptions {
--- a/lib/interfaces/swagger-document-options.interface.ts
+++ b/lib/interfaces/swagger-document-options.interface.ts
@@ -1,3 +1,15 @@
+export type OperationIdFactory =
+  | ((
+      controllerKey: string,
+      methodKey: string,
+      pathVersionKey?: string
+    ) => string)
+  | ((
+      controllerKey: string,
+      methodKey: string,
+      pathVersionKey?: string
+    ) => string);
+
 export interface SwaggerDocumentOptions {
   /**
    * List of modules to include in the specification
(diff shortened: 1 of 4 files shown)


lib/interfaces/swagger-document-options.interface.ts
```diff
@@ -1,3 +1,15 @@
+export type OperationIdFactory =
```
kamilmysliwiec, line 1: Why union here?

deongroenewald: Oh, not sure how that slipped in. Removed.

lib/interfaces/swagger-document-options.interface.ts
```diff
@@ -1,3 +1,9 @@
+export type OperationIdFactory = (
+  controllerKey: string,
+  methodKey: string,
+  pathVersionKey?: string
```
kamilmysliwiec, line 4:
```suggestion
  version?: string
```

lib/swagger-explorer.ts
```diff
@@ -275,6 +279,25 @@ export class SwaggerExplorer {
...
+      } else if (versionOrVersions !== VERSION_NEUTRAL) {
+        versions = [versionOrVersions];
+      }
+
+      const versionConfig = applicationConfig.getVersioning();
+      if (versionConfig?.type == VersioningType.URI) {
+        const prefix = this.routePathFactory.getVersionPrefix(versionConfig);
+        versions = versions.map((v) => `${prefix}${v}`);
+      }
+    }
```
kamilmysliwiec, line 299: should we move this logic to a dedicated method?

kamilmysliwiec:
LGTM

Issue #2615, opened 2023-09-14
Incorrect enum type always 'array'


alexandrulesi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When specifying `ApiPropertyOptions` enum the OAS now always now specifies the type to be 'array' instead of what it actually is

### Minimum reproduction code

https://github.com/alexandrulesi/nestjs-bug-repro

### Steps to reproduce

1. npm i
2. npm start
3. check console log with faulty type
4. change nestjs/swagger version to 7.1.8
5. npm i
6. check console log with correct type

### Expected behavior

Enum types should be what it is, in this case string, and not array

You can also check the json dumps for comparison

### Package version

7.1.11

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

kamilmysliwiec:
Would you like to create a PR for this issue?

alexandrulesi:
I'm afraid I'm not sure I'll find the time at the moment

IodizedGabe:
This appears to have been introduced by this change:
https://github.com/nestjs/swagger/pull/2592

Previously, this would always assume that the enum was a string. Now that it takes into account the declared type, it can erroneously be set as an array.

PR #2647, opened 2023-10-03
fix(): Derive enum types correctly when defined as an array #2615

IodizedGabe:
If an enum property is decorated with `isArray: true`, its type should be derived from the type of the enum item, rather than as an `array`.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/swagger/issues/2615

## What is the new behavior?
Enum schemas should now use the type of enum itself, regardless of whether it was declared as an array or not.

- [ ] Yes
- [x] No

### DIFF
--- a/test/services/schema-object-factory.spec.ts
+++ b/test/services/schema-object-factory.spec.ts
@@ -32,6 +32,12 @@ describe('SchemaObjectFactory', () => {
       Neighboard = 'neighboard'
     }
 
+    enum Ranking {
+      First = 1,
+      Second = 2,
+      Third = 3
+    }
+
     class CreatePersonDto {
       @ApiProperty()
       name: string;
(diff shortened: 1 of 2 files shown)


test/services/schema-object-factory.spec.ts
```diff
@@ -42,21 +48,36 @@ describe('SchemaObjectFactory', () => {
     class Person {
       @ApiProperty({ enum: Role, enumName: 'Role' })
       role: Role;
+
       @ApiProperty({ enum: Role, enumName: 'Role', isArray: true })
       roles: Role[];
```
IodizedGabe, line 53: A similar issue wasn't detected here with `roles` likely because `role` was processed first, where it isn't defined as an array.

test/services/schema-object-factory.spec.ts
```diff
@@ -42,21 +48,36 @@ describe('SchemaObjectFactory', () => {
...
+
+      @ApiProperty({ enum: Group, enumName: 'Group', isArray: true })
+      groups: Group[];
+
+      @ApiProperty({ enum: Ranking, enumName: 'Ranking', isArray: true })
+      rankings: Ranking[];
```
IodizedGabe, line 59: Also check that number-based enums also retain the correct type.

PR #2745, opened 2023-12-07
fix(core): support non-ASCII characters in filename paths

### LINKED ISSUE #1868: "Module not found" run-time error, if the project directory name contains non-ASCII characters

### Current behavior

I get following error when I start the server:
```
$ npm start                 
> server@0.0.1 start
> nest start

Error: Cannot find module '/home/xyz/Documents/Programozás/2022/project/src/AType'
Require stack:
- /home/xyz/Documents/Programozás/2022/project/dist/app.controller.js
- /home/xyz/Documents/Programozás/2022/project/app.module.js
- /home/xyz/Documents/Programozás/2022/project/main.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:101:18)
    at Object.<anonymous> (/home/xyz/Documents/Programozás/2022/project/dist/app.controller.js:31:46)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Module.require (internal/modules/cjs/loader.js:974:19)
```

### Minimum reproduction code

https://github.com/retnag/nestjs-swagger-apostrophe-in-path-error

### Steps to reproduce

1. Create a Directory, that contains a non-ASCII character in its name (I have tested it with "é" and "á". Example: "Programozás")
2. `cd` into it, and clone the [reproduction code](https://github.com/retnag/nestjs-swagger-apostrophe-in-path-error) into it. The goal here is, to have a non-ASCII character in the full path of the projects Filenames.
3. run `npm i` as usual
4. `npm start`

Key requirements provided by the [linked](https://github.com/retnag/nestjs-swagger-apostrophe-in-path-error) repo:
1. The full Path of where the project resides contains a non-ASCII character.
2. The [swagger cli-plugin](https://docs.nestjs.com/openapi/cli-plugin#using-the-cli-plugin) is enabled
3. The explicit or inferred return-type of a controller function, that handles an endpoint, is a class imported from another file in the project. 

### Expected behavior

I would expect the server to start without the error. I spent hours figuring out what's going on here, tests ran fine. Typescript compilation runs fine. I have tracked down the error to the following line in the emitted code by typescript (file: dist/app.controller.js:31): 
```
openapi.ApiResponse({ status: 201, type: require("/home/xyz/Documents/Programoz\u00E1s/2022/project/src/AType").AType })
```
Notice the \u00E1s Unicode escape sequence in the file path string. Now the "Module not found" error makes sense, since this directory really doesn't exist....

Also, and this might be a separate Issue altogether, I would expect a relative path here, as this absolute path is only going to be valid on the developer machine, ergo the build will fail with this error on any other machine It's deployed to.

### Package version

5.2.1

### NestJS version

8.0.0

### Node.js version

14.19.0

- [ ] macOS
- [X] Windows
- [X] Linux

### Other

This was successfully reproduced on another machine on Windows.

I guess this is not a high-priority issue, as a simple workaround would be not to use Unicode characters in your directory structure, but guys, its 2022 :'), things like this really shouldn't be an issue anymore, I just lost hours debugging this, as I never had suspected a user error, or a circular dependency somwhere. Please fix this.

BUT: the fact that there is a full, dev-environment relevant absolute path emittet there, which will break the server in a production environment, kinda seems like a dealbreaker to me!

Hannes-Kunnen:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If there is a non-ASCII character in the path of a dto file this will result in failure to normalize the path.

Example of a non-ASCII character in the project path (actual/expected):
```diff
- ./../../../../../../Priv/u00E9/forks/nestjs-swagger/test/plugin/fixtures/project/cats/dto/pagination-query.dto
+ ./cats/dto/pagination-query.dto
```
Example of a non-ASCII character in the dto folder (actual/expected):
```diff
- /home/user/Documents/projects/nestjs-swagger/test/plugin/fixtures/project/cats/dto/priv\u00E9/unicode.dto
+ ./cats/dto/privé/unicode.dto
```

This results in incorrect imports in the generated JavaScript.

Issue Number: Closes #1868

## What is the new behavior?

The swagger plugin now supports Unicode characters in paths to dto files.

- [ ] Yes
- [x] No

### DIFF
--- a/lib/plugin/utils/ast-utils.ts
+++ b/lib/plugin/utils/ast-utils.ts
@@ -103,7 +103,17 @@ export function getText(
     typeFormatFlags = getDefaultTypeFormatFlags(enclosingNode);
   }
   const compilerNode = !enclosingNode ? undefined : enclosingNode;
-  return typeChecker.typeToString(type, compilerNode, typeFormatFlags);
+
+  const text = typeChecker.typeToString(type, compilerNode, typeFormatFlags);
+
+  // Regex to match escaped unicode
+  const unicodeRegex = /\\u([\dA-Fa-f]{4})/g
+
+  // Unescape escaped unicode https://github.com/microsoft/TypeScript/issues/36174
+  return text.replace(unicodeRegex, (match, unicode) => {
+    const charCode = parseInt(unicode, 16);
+    return String.fromCharCode(charCode);
+  });
 }
 
 export function getDefaultTypeFormatFlags(enclosingNode: Node) {
--- a/test/plugin/fixtures/project/cats/dto/create-cat.dto.ts
+++ b/test/plugin/fixtures/project/cats/dto/create-cat.dto.ts
@@ -12,6 +12,7 @@ import { randomUUID } from 'node:crypto';
 import { ApiExtraModels, ApiProperty } from '../../../../lib';
 import { ExtraModel } from './extra-model.dto';
 import { LettersEnum } from './pagination-query.dto';
+import { UnicodeDto } from './privé/unicode.dto';
 import { TagDto } from './tag.dto';
 
 enum NonExportedEnum {
@@ -116,6 +117,9 @@ export class CreateCatDto {
 
   @ApiProperty({ description: 'tag', required: false })
   tag: TagDto;
+
+  @ApiProperty({ description: 'unicode', required: false })
+  unicode: UnicodeDto;
 
   multipleTags: TagDto[];
 
--- a/test/plugin/fixtures/project/cats/dto/privé/unicode.dto.ts
+++ b/test/plugin/fixtures/project/cats/dto/privé/unicode.dto.ts
@@ -1,0 +1,6 @@
+import { ApiProperty } from '../../../../../lib';
+
+export class UnicodeDto {
+  @ApiProperty({ description: 'name' })
+  name: string;
+}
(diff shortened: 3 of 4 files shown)


lib/plugin/utils/ast-utils.ts
```diff
@@ -103,7 +103,17 @@ export function getText(
...
   const compilerNode = !enclosingNode ? undefined : enclosingNode;
-  return typeChecker.typeToString(type, compilerNode, typeFormatFlags);
+
+  const text = typeChecker.typeToString(type, compilerNode, typeFormatFlags);
+
+  // Regex to match escaped unicode
+  const unicodeRegex = /\\u([\dA-Fa-f]{4})/g
+
+  // Unescape escaped unicode https://github.com/microsoft/TypeScript/issues/36174
+  return text.replace(unicodeRegex, (match, unicode) => {
```
kamilmysliwiec, line 113: I'm somewhat worried about the performance implications of this change. Are we sure we want to run this regex on every invocation of `getText`?

Issue #2723, opened 2023-11-21
Swagger documentation excludes any properties that uses the Matches class validator


abdelrahmansafwat:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

If there is a type that has a property validated with Match class-validator, that property is ignored by Swagger altogether.

### Minimum reproduction code

https://github.com/abdelrahmansafwat/matches-bug-repro

### Steps to reproduce

1. yarn install
2. yarn start
3. Navigate to http://localhost:3000/docs
4. Check CreateDemoDto in Schemas

### Expected behavior

The schema should be:
```
{
  name: string;
  age?: number;
}
```

### Package version

7.1.16

### NestJS version

10.0.0

### Node.js version

18.18.2

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

### Other

![284561800-20f03035-5a6b-4271-9da6-4732f2338101](https://github.com/nestjs/swagger/assets/33398534/d80f4d29-5e2c-426f-8d6d-5aabe2c32fc4)
![284561700-22dc9e81-a7cd-4994-9f19-f3818d9100d9](https://github.com/nestjs/swagger/assets/33398534/0f5688b9-b5fe-43c5-b735-847abe961e59)

quangtran88:
The root cause lies in the following line:
https://github.com/nestjs/swagger/blob/33bec34405a61910a463cf2927a7f2494b066f10/lib/plugin/visitors/model-class.visitor.ts#L673
When the first argument is a chained expression, `head(decoratorArguments).text` resolves to `undefined`. To address this issue, we can use the getText() method for a safer extraction of the pattern information passed in the first argument. For example:
- `@Matches(a.b.c)` will resolve as `"a.b.c"` (instead of `undefined` using `.text`)
- `@Matches(a)` will resolve as `"a"` (same as using `.text`)
- `@Matches(/^[+]?abc$/)` will resolve as `"/^[+]?abc$/"` (same as using `.text`)

kamilmysliwiec:
Let's track this here https://github.com/nestjs/swagger/pull/2747

PR #2747, opened 2023-12-08
fix(plugin): include property if using @Matches with chained expression

quangtran88:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When utilizing the `@Matches` decorator with a chained expression as the first argument, the associated property is not included in the `_OPENAPI_METADATA_FACTORY`. Consequently, this omission leads to the property being absent in the generated Swagger schema.

Issue Number: https://github.com/nestjs/swagger/issues/2723

## What is the new behavior?

What is the current behavior?
When utilizing the `@Matches` decorator with a chained expression as the first argument, the associated property is not included in the _OPENAPI_METADATA_FACTORY. Consequently, this omission leads to the property being absent in the generated Swagger schema.

Issue Number: https://github.com/nestjs/swagger/issues/2723

What is the new behavior?
The property decorated with `@Matches` with a chained expression as the first argument, should now be properly defined in the generated Swagger schema. This resolves the issue of missing properties in the Swagger documentation.

- [ ] Yes
- [x] No

## Other information
The root cause lies in the following line:
https://github.com/nestjs/swagger/blob/33bec34405a61910a463cf2927a7f2494b066f10/lib/plugin/visitors/model-class.visitor.ts#L673
When the first argument is a chained expression, `head(decoratorArguments).text` resolves to `undefined`. To address this issue, we can use the getText() method for a safer extraction of the pattern information passed in the first argument. For example:
- `@Matches(a.b.c)` will resolve as `"a.b.c"` (instead of `undefined` using `.text`)
- `@Matches(a)` will resolve as `"a"` (same as using `.text`)
- `@Matches(/^[+]?abc$/)` will resolve as `"/^[+]?abc$/"` (same as using `.text`)


micalevisk:
can you show an example of how the swagger UI would look like?

quangtran88:
> can you show an example of how the swagger UI would look like?

Given the model:
```ts
export const DemoValidation = {
  a: {
    b: /^abc$/,
    c: { d: /^abc$/ },
    e: [/^abc$/],
  },
};

const PATTERN = /^abc$/;

export class CreateDemoDto {
  @IsString()
  @Matches(DemoValidation.a.b)
  name: string;

  @IsString()
  @Matches(DemoValidation.a.c.d)
  name_1: string;

  @IsString()
  @Matches(DemoValidation.a.e[0])
  name_2: string;

  @IsString()
  @Matches(/^abc$/)
  name_3: string;

  @Matches(PATTERN)
  @IsOptional()
  name_4: string;
}
```
image
image

micalevisk:
I like the ideia but it also feels like we are leaking internal info

Can't expose the regex (ie, the pattern itself) instead of a 'label'? :thinking:

okonon:
Running into same issues that this PR fixes. Thanks for the all the work guys

okonon:
@micalevisk @quangtran88 are there any remaining to-dos to get this merged besides resolving conflicts ?

quangtran88:
I have successfully resolved the conflicts, thanks to @okonon. Upon further investigation, I discovered that this plugin generates Swagger metadata during the compilation of the source code. Consequently, obtaining the actual value at this stage is not feasible. I believe the focus for addressing this issue should be on fixing the missing property error. As for the new feature proposed by @micalevisk, it requires a more in-depth investigation and additional effort. It would be advisable to open a separate issue for this. What are your thoughts on this, @micalevisk?

micalevisk:
A new issue to report that seems good

test/plugin/fixtures/create-cat.dto.ts
```diff
@@ -90,7 +99,7 @@ export class CreateCatDto {
...
..., age: { required: true, type: () => Number, default: 3, minimum: 0, maximum: 10 }, positive: { required: true, type: () => Number, default: 5, minimum: 1 }, negative: { required: true, type: () => Number, default: -1, maximum: -1 }, lengthMin: { required: true, type: () => String, minLength: 2 }, lengthMinMax: { required: true, type: () => String, minLength: 3, maxLength: 5 }, tags: { required: true, type: () => [String] }, status: { required: true, default: Status.ENABLED, enum: Status }, status2: { required: false, enum: Status }, statusArr: { required: false, enum: Status, isArray: true }, oneValueEnum: { required: false, enum: OneValueEnum }, oneValueEnumArr: { required: false, enum: OneValueEnum }, breed: { required: false, type: () => String, title: "this is breed im comment" }, nodes: { required: true, type: () => [Object] }, optionalBoolean: { required: false, type: () => Boolean }, date: { required: true, type: () => Date }, twoDimensionPrimitives: { required: true, type: () => [[String]] }, twoDimensionNodes: { required: true, type: () => [[OtherNode]] }, cryptoUUIDProperty: { required: true, type: () => String }, arrayOfUUIDs: { required: true, type: () => [String] } };
```
kamilmysliwiec, line 102: `pattern: "Pattern.isNumber"` with this, the `Pattern.isNumber` string would be shown in the UI (not /^d+$/)

PR #2757, opened 2023-12-25
Move typescript from dev to peer dependencies.

sunoru:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The current version lists `typescript` as a dev dependency, which means it wouldn't be installed with the package in production. However, this package directly depends on importing the typescript library like

https://github.com/nestjs/swagger/blob/18f088e39ef586c8af6d4efbdaf4bf3bdaffc556/lib/plugin/visitors/abstract.visitor.ts#L1

Since all nestjs projects are already in typescript, and `@nestjs/swagger` is only used in development, it is usually not an issue. However, if people are using newer package managers like pnpm or yarn+pnp, packages not declared under `dependencies` or `peerDependencies` will be strictly forbidden to import. It will cause problems like https://github.com/nestjs/nest-cli/issues/1836

## What is the new behavior?

`typescript` is now listed as a peer dependency, which solves the problem mentioned above.a

- [ ] Yes
- [x] No

It shouldn't introduce a breaking change since it's only moving a dependency from dev to peer.

## Other information

I didn't write any tests in this PR. I don't know if they are necessary for this small change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,8 +44,7 @@
         "release-it": "17.0.1",
         "supertest": "6.3.3",
         "swagger-parser": "10.0.3",
-        "ts-jest": "29.1.1",
-        "typescript": "5.1.6"
+        "ts-jest": "29.1.1"
       },
       "peerDependencies": {
         "@fastify/static": "^6.0.0",
@@ -53,7 +52,8 @@
         "@nestjs/core": "^9.0.0 || ^10.0.0",
         "class-transformer": "*",
         "class-validator": "*",
-        "reflect-metadata": "^0.1.12"
+        "reflect-metadata": "^0.1.12",
+        "typescript": "5.1.6"
       },
       "peerDependenciesMeta": {
         "@fastify/static": {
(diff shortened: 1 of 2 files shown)


micalevisk:
> Since all nestjs projects are already in typescript

we can use nestjs with plain js. See https://github.com/nestjs/javascript-starter/

> and @nestjs/swagger is only used in development

that not true when we import anything from `@nestjs/swagger` (like those `@Api*` decorators)

sunoru:
That’s interesting, thanks for the explanation. Then maybe the best solution should be splitting `@nestjs/swagger` and the CLI plugin it provides into two packages. Do I understand correctly?

kamilmysliwiec:
Thank you for taking the time to submit your report! From the looks of it, this could be better discussed on our Discord. If you haven't already, please join [here](https://discord.gg/nestjs) and send a new post in the #⁠ 🐈 nestjs-help forum. Make sure to include a link to this issue, so you don't need to write it all again. We have a large community of helpful members, who will assist you in getting this to work.

PR #2797, opened 2024-01-25
fix: support type `UUID` from node's `crypto` package

TheSalarKhan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Using the `UUID` type from node's `crypto` package with the `@nestjs/swagger` plugin enabled crashes nestjs start `npm run start` with a cryptic error message.

```
Error: A circular dependency has been detected (property key: "id"). Please, make sure that each side of a bidirectional relationships are using lazy resolvers ("type: () => ClassType").
```

Example dto
```
import { UUID } from 'crypto';

export class CreateUserDto {
  name: string;

  id: UUID;
}
```

This problem goes away when the `IsUUID()` decorator from `class-validator` package is used. But considering the fact that UUID is an alias ```type UUID = `${string}-${string}-${string}-${string}-${string}`;``` the inferred type without any help should be `string`

Issue Number: N/A

## What is the new behavior?

After this change, if the plugin is enabled, and no decorator is used on a property of type `crypto.UUID` then its type gets inferred as `string` in the openapi spec.

- [ ] Yes
- [x] No

### DIFF
--- a/lib/plugin/utils/plugin-utils.ts
+++ b/lib/plugin/utils/plugin-utils.ts
@@ -115,6 +115,9 @@ export function getTypeReferenceAsString(
     if (type.aliasSymbol) {
       return { typeName: 'Object', arrayDepth };
     }
+    if (typeChecker.getApparentType(type).getSymbol().getEscapedName() === 'String') {
+      return { typeName: String.name, arrayDepth };
+    }
     return { typeName: undefined };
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #2872, opened 2024-03-11
fix(plugin): call assignment function to add `enum` property by `@IsEnum` decorator

seonggukchoi:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When the option `classValidatorShim` is enabled, I expected the CLI to add an `enum` property into Swagger documentation automatically about a `@IsEnum` decorator used property, but it doesn't run expectation.
Then I tried using a `@IsIn` decorator instead of `@IsEnum`, and the behavior ran as expected.

Issue Number: N/A

## What is the new behavior?

When the option `classValidatorShim` is enabled and using the `@IsEnum` decorator at a property, the CLI plugin adds the `enum` property automatically.

- [ ] Yes
- [x] No

## Other information

If I need then I will create an issue with the information, I don't know why exists the `readonly` option, and why can't run adding `enum` options by the `@IsIn` decorator when enabled the option `readonly`.

### DIFF
--- a/lib/plugin/visitors/model-class.visitor.ts
+++ b/lib/plugin/visitors/model-class.visitor.ts
@@ -640,6 +640,14 @@ export class ModelClassVisitor extends AbstractFileVisitor {
         assignments,
         options
       );
+      this.addPropertyByValidationDecorator(
+        factory,
+        'IsEnum',
+        'enum',
+        decorators,
+        assignments,
+        options
+      );
     }
     this.addPropertyByValidationDecorator(
       factory,
--- a/lib/plugin/visitors/model-class.visitor.ts
+++ b/lib/plugin/visitors/model-class.visitor.ts
@@ -631,7 +631,7 @@ export class ModelClassVisitor extends AbstractFileVisitor {
     const decorators = ts.canHaveDecorators(node) && ts.getDecorators(node);
 
     if (!options.readonly) {
-      // @IsIn() and @IsEnum() decorator is not supported in readonly mode
+      // @IsIn() and @IsEnum() decorators are not supported in readonly mode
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Could you update tests too?

lib/plugin/visitors/model-class.visitor.ts
```diff
@@ -631,7 +631,7 @@ export class ModelClassVisitor extends AbstractFileVisitor {
     const decorators = ts.canHaveDecorators(node) && ts.getDecorators(node);
 
     if (!options.readonly) {
-      // @IsIn() annotation is not supported in readonly mode
+      // @IsIn() and @IsEnum() decorator is not supported in readonly mode
```
aqeelat, line 634:
```suggestion
      // @IsIn() and @IsEnum() decorators are not supported in readonly mode
```

seonggukchoi:
Hi, @kamilmysliwiec. I haven't been able to pay attention to this PR recently.
I have read some of the test files that seem to be related to the `model-class.visitor.ts` file, but I couldn't intuitively understand them.
So, I'm not sure which files to add the tests to and how to add them.
Could you please help me add tests? 
If there exists any guide on writing tests in this project, documentation explaining the module structure, similar tests, or anything else would be helpful.

kamilmysliwiec:
> I expected the CLI to add an enum property into Swagger documentation automatically about a @IsEnum decorator used property, but it doesn't run expectation

This is the expected behavior; we automatically include the enum attribute in other situations, specifically when the property type is an actual enum.

Issue #3160, opened 2024-11-11
@ApiSchema not working as expected when inheriting


arabold:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Assume you define two DTOs that inherit from each other:
```ts
@ApiSchema({ name: 'BaseResponse' })
class BaseResponseDto {
}

@ApiSchema({ name: 'Response' })
class ResponseDto extends BaseResponse {
}
```

Then use this class in your controller:
```ts
@Controller()
class MyController() {
   @Get()
   @ApiOkResponse({ type: ResponseDto })
   async doSomething(): Promise<ResponseDto> {
     // ...
   }
}
```

In the resulting OpenAPI spec you would expect that the response schema for `doSomething` is `Reponse`, the specialized subclass `ResponseDto`. However, the OpenAPI spec uses `BaseResponse` instead, the parent class.

### Minimum reproduction code

Example in the description above

### Expected behavior

For inherited classes I would expect the schema name of the subclass to be used, not of the parent class.

The error lies in the following piece of code in [lib/services/schema-object-factory.ts](https://github.com/nestjs/swagger/blob/7acedba5c8738a75e54c209c60173ea57639e99a/lib/services/schema-object-factory.ts#L295):

```ts
    const schemaName = customSchema[0].name;
```

It should be replaced by

```ts
    const schemaName = customSchema[customSchema.length - 1].name;
```

to use the schema name of the subclass.

### Package version

8.0.5

### NestJS version

10.4.7

### Node.js version

20.11.1

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

kamilmysliwiec:
Would you like to create a PR for this issue?

PR #3164, opened 2024-11-13
fix: @ApiSchema inheritance

arabold:
Ensure the correct @ApiSchema name is used when inheriting from a base class.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #3160

## What is the new behavior?

The correct schema name will be used

- [ ] Yes
- [x] No

## Other information

There's also a fix necessary for `getSchemaPath` (get-schema-ath.util.ts). However I have not found a single test for these utils, so I didn't add one but applied the exact same fix.

### DIFF
--- a/test/services/schema-object-factory.spec.ts
+++ b/test/services/schema-object-factory.spec.ts
@@ -412,6 +412,24 @@ describe('SchemaObjectFactory', () => {
       expect(Object.keys(schemas)).toContain('UpdateUserDto');
     });
 
+    it('should override the schema name of base class', () => {
+      @ApiSchema({
+        name: 'CreateUser'
+      })
+      class CreateUserDto {}
+
+      @ApiSchema({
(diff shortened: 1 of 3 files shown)


test/services/schema-object-factory.spec.ts
```diff
@@ -412,6 +412,24 @@ describe('SchemaObjectFactory', () => {
...
 
+    it('should override the schema name of base class', () => {
+      @ApiSchema({
+        name: 'CreateUser'
+      })
+      class CreateUserDto {}
+
+      @ApiSchema({
+        name: 'UpdateUser'
+      })
```
arabold, line 423: Interesting enough, the case _without_ an `@ApiSchema` decorator in the child class has been tested. Just not the case when both the parent and the child class have a decorator set.

Issue #3102, opened 2024-10-14
[discussion]: allow swagger-scanner deepScanRoutes to be recursive or go down an extra level


raphaelroshan:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Swagger scanner does not support going more than one level down (with the deepScanRoutes option), and I would like for there to be an additional flag/ parameter that would allow it to scan 2 levels down, or recursively if possible.

### Describe the solution you'd like

I would like to implement a recursiveScan flag that would allow Swagger to scan more than one level down, in the case of such consolidated apps. If this is against design decision, perhaps I would adapt the deepScan to accept a parameter of 1 (default) or 2, scanning an extra level down. This change would be implemented entirely in swagger-scanner.ts.

### Teachability, documentation, adoption, migration strategy

I will adapt any existing swagger-scanner documentation and update it for this new change.

### What is the motivation / use case for changing the behavior?

Swagger was a functional choice when we operated several microservices, as we only needed to deepScan one level down. Recently we have consolidated several microservices into an app, and as a result we need to scan an additional level down, which swagger does not currently support. To comply with the single level deepScan would require shifting all the routes up one level which would mess up the way the app is configured as a consolidation of sub-apps.

raphaelroshan:
Yes, I will submit it next week. Anything to note?

evgenyidf:
Any update ?

raphaelroshan:
Still working on it! Will submit a PR in a few days hopefully 🙏

hee-c:
I made [PR](https://github.com/nestjs/swagger/pull/3186) to this issue.

raphaelroshan:
Aw dang was still working on mine, but since I couldnt get it to work probably would be better to go with yours! Had a question if anybody here could help me out answering it - this was my updated code for denormalized paths:
```
const denormalizedPaths = modules.map(
        ({ controllers, metatype, imports }) => {
          let result: ModuleRoute[] = [];
          const isGlobal = (module: Type<any>) => !container.isGlobalModule(module);
          if (deepScanRoutes || recursiveScanRoutes) {
            const scanSubmodules = (module) => {
              if (!isGlobal(module)) {
                module.imports.forEach(subModule => {
                  const modulePath = this.getModulePathMetadata(container, subModule.metatype);
                  result = result.concat(
                    this.scanModuleControllers(
                      subModule.controllers,
                      modulePath,
[...]

PR #3186, opened 2024-11-24
feat: Add recursiveModuleScan option to scanner

hee-c:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The Swagger scanner only scans up to 1-depth for modules included via the `include` option, even when `deepScanRoutes` is set to `true`.

Issue Number: #3102

## What is the new behavior?
스크린샷 2024-11-24 오후 2 01 00

When `deepScanRoutes` is set to `true` and the newly added `recursiveModuleScan` flag is also set to `true`, the modules included via `include` will be scanned recursively up to the depth defined by `maxScanDepth`.

- **`deepScanRoutes: false`**: Only scans root modules imported by `include` (current behavior).
- **`deepScanRoutes: true`**: Scans 1-depth modules imported by `include` (current behavior).
- **`deepScanRoutes: true, recursiveModuleScan: true`**: Scans all modules recursively, with no depth limit.
- **`deepScanRoutes: true, recursiveModuleScan: true, maxScanDepth: 5`**: Scans modules recursively up to a depth of 5.

- [ ] Yes
- [x] No

### DIFF
--- a/e2e/src/dogs/depth1-dogs/depth1-dogs.controller.ts
+++ b/e2e/src/dogs/depth1-dogs/depth1-dogs.controller.ts
@@ -1,0 +1,11 @@
+import { Controller, Get } from '@nestjs/common';
+import { ApiTags } from '../../../../lib';
+
+@ApiTags('depth1-dogs')
+@Controller('depth1-dogs')
+export class Depth1DogsController {
+  @Get()
+  findAll() {
+    return 'Depth1 Dogs';
+  }
+}
--- a/e2e/validate-schema.e2e-spec.ts
+++ b/e2e/validate-schema.e2e-spec.ts
@@ -13,6 +13,7 @@ import {
 import { ApplicationModule } from './src/app.module';
 import { Cat } from './src/cats/classes/cat.class';
 import { TagDto } from './src/cats/dto/tag.dto';
+import { DogsModule } from './src/dogs/dogs.module';
 
 describe('Validate OpenAPI schema', () => {
   let app: INestApplication;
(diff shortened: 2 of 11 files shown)


e2e/src/dogs/depth1-dogs/depth1-dogs.controller.ts
```diff
@@ -0,0 +1,11 @@
+import { Controller, Get } from '@nestjs/common';
+import { ApiTags } from '../../../../lib';
+
+@ApiTags('depth1-dogs')
+@Controller('depth1-dogs')
+export class Depth1DogsController {
```
hee-c, line 6: The files under `e2e/src/dogs` are modules used for testing purposes.

e2e/validate-schema.e2e-spec.ts
```diff
@@ -215,3 +216,115 @@ describe('Validate OpenAPI schema', () => {
     });
   });
 });
+
+describe('Nested module scanning', () => {
```
hee-c, line 220: I have added the tests here for now, but if you prefer to separate them into a dedicated file, let me know and I will update accordingly.

lib/interfaces/swagger-document-options.interface.ts
```diff
@@ -53,4 +53,20 @@ export interface SwaggerDocumentOptions {
...
+   */
+  recursiveModuleScan?: boolean;
+
+  /**
+   * Maximum depth level for recursive module scanning.
+   * Only applies when `recursiveModuleScan` is `true`.
+   *
+   * @default Infinity
+   */
+  maxScanDepth?: number;
```
hee-c, line 71: The default value for `maxScanDepth ` is currently set to `Infinity`. If you have any suggestions, let me know.

lib/swagger-scanner.ts
```diff
@@ -170,4 +193,71 @@ export class SwaggerScanner {
...
+          );
+        }
+      }
+    }
+
+    return result;
+  }
+
+  private getModuleId(metatype: Type<any>): string {
+    return metatype.name || Math.random().toString(36).substring(2);
```
kamilmysliwiec, line 261: Not sure if I understand what's the purpose of having `Math.random().toString(36).substring(2)` if it isn't calculated based on the input argument value (metatype)

hee-c: You're right about the random fallback being unnecessary. Let me explain the context better. The `getModuleId` is used with `processedModules` set to prevent duplicate scanning and circular dependencies while recursively scanning module imports. However, I realized that using just `metatype.name` could be problematic - if we have two different modules with the same name in different locations (like auth.module.ts in different directories), only one would be scanned. To properly handle such cases while still preventing infinite recursion, I think we should use `metatype.toString()`. While this generates longer strings, it ensures we don't miss any modules during the scanning process. What do you think about this approach?
```suggestion
    return metatype.toString();
```

kamilmysliwiec: Sounds reasonable. I'm afraid the size of the array might become notably large for bigger projects though

hee-c: Looking at `ModuleTokenFactory.create`, I see that `id` uses uid(21) and `token` adds metatype.name and dynamicModuleMetadata(optional) to that. Since `NestContainer.addModule` seems to use `token` as the identifier, rather than using metatype, what if we destructure `token` along with the existing properties in the `imports.values()` loop within the `scanModuleImportsRecursively` method and use that as the moduleId?

lib/swagger-scanner.ts
```diff
@@ -170,4 +193,71 @@ export class SwaggerScanner {
...
+      ) => string;
+      autoTagControllers?: boolean;
+    }
+  ): ModuleRoute[] {
+    let result: ModuleRoute[] = [];
+
+    for (const { metatype, controllers, imports: subImports } of imports.values()) {
+      // Skip if module has already been processed
+      const moduleId = this.getModuleId(metatype);
+      if (processedModules.has(moduleId) || container.isGlobalModule(metatype) || (maxDepth !== undefined && currentDepth > maxDepth)) {
```
hee-c, line 220:
```suggestion
    for (const { token, metatype, controllers, imports: subImports } of imports.values()) {
      // Skip if module has already been processed
      if (processedModules.has(token) || container.isGlobalModule(metatype) || (maxDepth !== undefined && currentDepth > maxDepth)) {
```
Here's an example implementation of the approach I suggested

Issue #2990, opened 2024-06-26
Support for Header Versioning


dalaverd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

The issue described in [nestjs/swagger#2235](https://github.com/nestjs/swagger/issues/2235) persists in the latest version. The Swagger module does not support the type of versioning with HEADER that NestJS supports in its core. There was a fix proposed 1 year ago ([PR #2063](https://github.com/nestjs/swagger/pull/2063)), but it has never been merged.

### Minimum reproduction code

https://github.com/DCampagnola/nestjs-swagger-header-versioning-bug

### Expected behavior

Generated api should include every version declared with the different docs

### Package version

7.3.1

### NestJS version

10.3.9

### Node.js version

20.14.0

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [ ] Linux

kamilmysliwiec:
Let's track this here https://github.com/nestjs/swagger/pull/3209

PR #3209, opened 2024-12-12
Fix/Feat: Support header-based versioning and filtering routes by version

aarongoin:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables) (formatting changes from running `npm run format`)
- [ ] Build related changes
- [ ] CI related changes

Currently `@nestjs/wagger` does not support header-based versioning that is documented as being supported here: https://docs.nestjs.com/techniques/versioning#header-versioning-type

In addition end users lack any ability to filter the routes documented in the OpenAPI specification down to some subset of versions their app supports.

Issue Number: #2990

## What is the new behavior?
Apps that use header-based versioning for endpoints (and I believe media-type based versioning as well though I haven't tested) are now properly supported. Endpoints of different versions using the same path no longer conflict with each other and are documented as separate endpoints in the resulting OpenAPI specification.

Additionally users can filter the endpoints documented down to a subset to better support use cases where there is a separate swagger document per API version.

- [ ] Yes
- [x] No

## Other information


e2e/header-version.e2e-spec.ts
```diff
@@ -0,0 +1,174 @@
...
+      api.info.title,
+      api.info.version
+    );
+    expect(api.info.title).toEqual('Dogs example');
+
+    expect(api.paths['/api/dogs']['get']).toBeDefined();
+
+    expect(api.paths['/api/dogs version: v0']).toBeUndefined();
+
+    expect(api.paths['/api/dogs version: v1']['post']['operationId']).toBe('DogsController_createNewV1');
```
kamilmysliwiec, line 145: Not sure if I'm following. How's this a valid path? (`/api/dogs version: v1`)

aarongoin: Apologies for the delayed response.

You're correct that it's not a valid path, but there doesn't seem to be any other way of distinguishing between different versioned paths that route via headers when generating OpenAPI spec covering multiple versions simultaneously--something thats possible when versions are specified in the path. Unless perhaps your suggesting to make it valid path via something like query params--though that doesn't seem ideal either.

I suppose the only other alternative is to force the user to specify a single API version to generate OpenAPI spec for (with option of including/excluding version-less routes). Do you prefer this alternative? Or do you have some other idea in mind for this?

PR #3219, opened 2024-12-26
refactor: Replace express/fastify specific methods with adapter methods

Makkalay:
Replace express/fastify specific methods with adapter methods

## PR Checklist

- [x ] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

req.send and erq.type is express/fastify specific code https://nodejs.org/api/http.html#requestenddata-encoding-callback

Issue Number: N/A

## What is the new behavior?
To work with any adapters.

- [ x] Yes
- [ ] No

Need to add that methods (send and type) to all adapters if not (official and custom).

### DIFF
--- a/lib/swagger-module.ts
+++ b/lib/swagger-module.ts
@@ -149,7 +149,7 @@ export class SwaggerModule {
     httpAdapter.get(
       normalizeRelPath(`${finalPath}/swagger-ui-init.js`),
       (req, res) => {
-        res.type('application/javascript');
+        httpAdapter.type(res, 'application/javascript');
         const document = getBuiltDocument();
 
         if (swaggerOptions.patchDocumentOnRequest) {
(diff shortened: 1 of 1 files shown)


micalevisk:
as you can see at https://github.com/nestjs/nest/blob/b6ea2a1899fe54f289e2c6188c843705c9072698/packages/core/adapters/http-adapter.ts#L12 we don't have `type` and `send` methods on the abstract http adapter

Instead of `res.send`, I think we can use `httpAdapter.reply`. And, if I'm not wrong, we can drop the `res.type` calls because `httpAdapter.reply` is supposed to define the Content-Type HTTP header as well. (see https://github.com/nestjs/nest/blob/b6ea2a1899fe54f289e2c6188c843705c9072698/packages/platform-fastify/adapters/fastify-adapter.ts#L336 and https://github.com/nestjs/nest/blob/b6ea2a1899fe54f289e2c6188c843705c9072698/packages/platform-express/adapters/express-adapter.ts#L65)

Makkalay:
No import, because not abstract class AbstractHttpAdapter will do the work but adapter, and in adapter we at least can name thing as we want. In express adapter reply call send.

micalevisk:
Sorry but I didn't follow what you meant

We can merge this PR because the tests are failing. But I agree that we shouldn't be using `res.send` and `res.type` in there.

Makkalay:
https://github.com/expressjs/express/blob/b11122be85377375465b2aa103da43301e56f128/lib/response.js#L113

Issue #3207, opened 2024-12-12
Array schemas are not generated properly


ccaspers:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When an array is declared, depending on where its used and how it was decorated, the output doesn't look as one would expect.

- the array type is sometimes not rendered correctly
- restrictions on array-length are missing 
- restrictions on the length of strings in a string array are missing 

### Minimum reproduction code

https://github.com/ccaspers/nestjs-swagger/commit/3248507c5c40abc7623cdee6663a50fc1e68bf1e

### Steps to reproduce

1. `git clone git@github.com:ccaspers/nestjs-swagger.git`
2. `cd nestjs-swagger`
3. `git switch array-schemas-bug`
4. `npm i`
5. `npm run test`

### Expected behavior

all tests succeed

### Package version

8.1.0

### NestJS version

10.4.15

### Node.js version

20.12.0

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

### Other

Maybe the syntax used in some of the decorators is wrong? But there is no documentation with regards to how things should not be declared.

Also, at least in the case for dtos for query parameters, there is no combination of syntax available that leads to the desired result.

kamilmysliwiec:
Would you like to create a PR for this issue?

kamilmysliwiec:
They look reasonable yes

mag123c:
@ccaspers Hello, If you are not working on it, can I create this PR?

kamilmysliwiec:
https://github.com/nestjs/swagger/pull/3218

ccaspers:
I checked #3218 but it will not work as expected. I have something that works and that makes the unit tests I created succeed. The changes feel extremely hacky right now. 

I'll try to improve on this when I find the time. Current state of my changes can be checked here
https://github.com/nestjs/swagger/compare/master...ccaspers:nestjs-swagger:array-schemas-bug

ccaspers:
@kamilmysliwiec maybe it makes sense to reopen this issue, since it is related to #3218 but not identical

jdespatis:
@ccaspers I experience the same problem as you,

for example in a dto containing:

```typescript
  @MinLength(3, { each: true })
  @IsString({ each: true })
  values: string[];
```

The NestJS Swagger plugin generates this:

```yaml
        - name: values
          required: false
          in: query
          schema:
            minLength: 3
            type: array
            items:
              type: string
````

[...]

PR #3255, opened 2025-01-18
fix(swagger): improve rendering of array items

ccaspers:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When documenting items in arrays, `minLength` and `maxLength` will not be rendered. 

Issue Number: #3207 

## What is the new behavior?

`minLength` and `maxLength` can now be used to document items in arrays.

- [ ] Yes
- [x] No

### DIFF
--- a/lib/services/swagger-types-mapper.ts
+++ b/lib/services/swagger-types-mapper.ts
@@ -115,8 +115,8 @@ export class SwaggerTypesMapper {
         ...omit(this.getSchemaOptions(param), [...itemsModifierKeys]),
         type: 'array',
         items: isString((items as any).type)
-          ? { type: (items as any).type, ...modifierProperties }
-          : { ...(items as any).type, ...modifierProperties }
+          ? { type: (items as any).type, ...modifierProperties, ...items }
+          : { ...(items as any).type, ...modifierProperties, ...items }
       }
     };
   }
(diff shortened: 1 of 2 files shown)


lib/services/swagger-types-mapper.ts
```diff
@@ -115,8 +115,8 @@ export class SwaggerTypesMapper {
         ...omit(this.getSchemaOptions(param), [...itemsModifierKeys]),
         type: 'array',
         items: isString((items as any).type)
-          ? { type: (items as any).type, ...modifierProperties }
-          : { ...(items as any).type, ...modifierProperties }
+          ? { type: (items as any).type, ...modifierProperties, ...items }
+          : { ...(items as any).type, ...modifierProperties, ...items }
```
kamilmysliwiec, line 119: `...items` will override `type` here (when type is not string)

ccaspers: That's true. But this will only happen, if the user decided to manually add a type to items like this for example:

```ts
@ApiProperty({
  items: {
    type: 'string',
  }
})
```

I would argue that this is the desired behavior, but I'll look into it.

kamilmysliwiec: I have seen this fairly often in the past

PR #3353, opened 2025-03-21
chore(swagger): include model name in circular dependency warning

dunklesToast:
## PR Checklist

## PR Type

- [x] Other - Improved Error Message

Right now nest shows something like this:
`Error: A circular dependency has been detected (property key: "id"). Please, make sure that each side of a bidirectional relationships are using lazy resolvers ("type: () => ClassType").`. If you've changed / refactored few files in your repo you probably don't easily know with Model had the error (and since the error occurred in the `id` field for me, it could've been basically any)

## What is the new behavior?
We're now checking if there is a `pendingSchemaRef`. As far as I understood the code this is a list of models to work on. In my testing the first item always equaled the model that the error was thrown so I thought it might be a good idea to include the model name in the error message.
If I was just "lucky" and the order of the `pendingSchemaRefs` isn't actually sorted in any way making the error message invalid, I guess we can close this (if there is no other easy way to infer the models name)

- [x] No

### DIFF
--- a/lib/services/schema-object-factory.ts
+++ b/lib/services/schema-object-factory.ts
@@ -453,7 +453,7 @@ export class SchemaObjectFactory {
   ): SchemaObjectMetadata {
     if (isUndefined(trueMetadataType)) {
       throw new Error(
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Yeah we can't rely on the `pendingSchemaRefs` order in this case

PR #3543, opened 2025-08-18
feat(decorators): accept RegExp in ApiProperty({ pattern }) while keeping schema pattern string

### LINKED ISSUE #3374: Allow to pass RegExp to @ApiProperty({ pattern })

As a practical development, to use a RegExp it is common that it is itself except in some global constant of the solution and then it is used in all applications.

The problem is that to export this RegExp, @ApiProperty({ pattern }) does not need to be converted to a string and then remove the initial and final bars as well.

```ts
// src/domain/constants/regex.ts
export const REGEX = {
  ALIAS: /^[a-z0-9]+(-[a-z0-9]+)*$|^.{5,50}$/
}

// src/resources/organization/dtos/create-organization.dto.ts
import {ApiProperty} from '@nestjs/swagger';
import {REGEX} from '#/domain/constants';

export class CreateOrganizationDTO {
  @ApiProperty({ pattern: pattern: REGEX.ALIAS.toString().slice(1, -1), })
  alias: string
}
```

It is a common behavior to specify converting the RegExp to string and removing the initial and final bars, and, as it is also possible to identify the type of the value attributed in a pattern using the `options.pattern instanceof RegExp`, I believe that it is simple to implement the behavior of using RegExp in this field. A possible solution is a simple addition [here](https://github.com/nestjs/swagger/blob/master/lib/decorators/api-property.decorator.ts#L61):

```ts
// lib/decorators/api-property.decorator.ts

export function createApiPropertyDecorator(
  options: ApiPropertyOptions = {},
  overrideExisting = true
): PropertyDecorator {
  if(options.pattern instanceof RegExp) {
    options.pattern = options.pattern.toString().slice(1, -1);
  }

  const [type, isArray] = getTypeIsArrayTuple(options.type, options.isArray);

  // ... rest of code here
```

The developers will be define properties passing the pattern in ApiProperty decorator as RegExp directly without adjust like:

```ts
import {ApiProperty} from '@nestjs/swagger';

export class ExampleDTO {
  @ApiProperty({ pattern: /^\w+$/ })
  example: string
}
```

Simplify use an coding experience of `@nestjs/swagger` package to developers

Kimxxunu:
Allow passing a RegExp to `ApiProperty({ pattern })`. The decorator normalizes the RegExp to an
OpenAPI-compatible string by stripping slashes and flags (e.g., `/^\w+$/gi` → `^\\w+$`). This keeps
the generated schema compliant with OpenAPI/JSON Schema where `pattern` must be a string.

- Input DX: `pattern?: string | RegExp` (options only)
- Output schema: `pattern` is always a string
- No breaking changes
- Added unit tests (5 cases) for normalization and immutability

Closes #3374

## PR Checklist
- [ ] Docs have been added / updated (not required for this change)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

`ApiProperty({ pattern })` only accepts a string. When developers want to reuse an app-wide `RegExp`,
they must convert it to a string and manually strip the leading/trailing slashes and ignore flags,
which is repetitive and error-prone.

Issue Number: #3374

## What is the new behavior?
`ApiProperty({ pattern })` now accepts `string | RegExp`. When a `RegExp` is provided, the decorator
normalizes it to a plain string (no slashes/flags) before storing metadata, keeping the generated
OpenAPI schema valid.

- [ ] Yes
- [x] No

## Other information
- Implementation localized to the decorator options processing:
  - `lib/decorators/api-property.decorator.ts`: accept `pattern?: string | RegExp` at options level,
    normalize via `RegExp#toString()` and remove slashes/flags before emitting metadata.
  - Does **not** change OpenAPI types: schema `pattern` remains a **string**.
  - Does not mutate the caller’s options object.

- Tests:
  - `test/decorators/api-property.decorator.spec.ts`
    - converts RegExp → string (removing slashes/flags)
    - keeps string pattern as-is
    - supports `new RegExp(...)`
    - preserves escaped slashes inside the body
    - does not mutate the original options object

### DIFF
--- a/lib/decorators/api-property.decorator.ts
+++ b/lib/decorators/api-property.decorator.ts
@@ -7,9 +7,18 @@ import {
 } from '../interfaces/schema-object-metadata.interface';
 import { getEnumType, getEnumValues } from '../utils/enum.utils';
 import { createPropertyDecorator, getTypeIsArrayTuple } from './helpers';
+import {
+  ReferenceObject,
+  SchemaObject
+} from '../interfaces/open-api-spec.interface';
 
-export type ApiPropertyCommonOptions = SchemaObjectMetadata & {
+type ApiPropertyCommonOptions = (Omit<SchemaObjectMetadata, 'pattern'> & {
+  pattern?: string | RegExp;
+  properties?: Record<string, SchemaObject | ReferenceObject>;
+  selfRequired?: boolean;
+}) & {
   'x-enumNames'?: string[];
+  selfRequired?: boolean;
   /**
    * Lazy function returning the type for which the decorated property
    * can be used as an id
@@ -38,6 +47,10 @@ const isEnumArray = (
   items: any;
 } => opts.isArray && 'enum' in opts;
 
+function normalizePattern(p?: string | RegExp): string | undefined {
+  return p instanceof RegExp ? p.source : p;
+}
+
 /**
  * @publicApi
  */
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
https://github.com/nestjs/swagger/pull/3787

Issue #3519, opened 2025-07-17
swagger plugin cause circular dependency when having a field of optional and nullable enum


boompikachu:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

When using @nestjs/swagger plugin to create a DTO, if the field is an optional nullable enum, the error [below] occurs

```
Error: A circular dependency has been detected (property key: "status"). Please, make sure that each side of a bidirectional relationships are using lazy resolvers ("type: () => ClassType").
```

From the repo, we see that `statusOne` and `statusTwo` works perfectly fine. But if we introduce `statusThree`, the one with optional and nullable, it cause the swagger to crash (circular dependency error)

```ts
export class CreateTaskDto {
  statusOne?: Status;
  statusTwo: Status | null;
  statusThree?: Status | null; // This cause circular dependency error when opening http://localhost:3000/api (swagger)

  @ApiProperty({ enum: Status, nullable: true })
  statusFour?: Status | null; // This works because @ApiProperty will override the swagger plugin
}
```

### Minimum reproduction code

https://github.com/boompikachu/nestjs-swagger-plugin-repro

### Steps to reproduce

1. `pnpm install`
2. `pnpm start:dev`
3. open http://localhost:3000/api

comment out the `statusThree?: Status | null` to fix the error

### Expected behavior

To not cause circular dependency

### Package version

11.2.0

### NestJS version

11.1.4

### Node.js version

24

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

in https://github.com/nestjs/swagger/blob/9fe5c1ff43d0ff702d73fb51632909e163557ba2/test/plugin/fixtures/nullable.dto.ts#L11, there's no test case for nullable and optional enum

RoiEf:
This seem to be the same as issue #3326

mmiglioranza22:
I've seen this issue mentioned in #1475 and in stack overflow [5 years ago](https://stackoverflow.com/questions/60749439/circular-dependency-with-nestjs-swagger-4). Seems that the case for Enums might be fixed in #3544 , however for plain `null` properties the message still refers to "circular dependencies".

I understand it surely is a much harder fix than just changing the error message, however mentioning this common pitfall somewhere in the docs would be of great help.

null prop fails
<img width="1111" height="574" alt="Image" src="https://github.com/user-attachments/assets/52729d47-e11d-4d0e-b054-9234434ef4ca" />
[...]

PR #3544, opened 2025-08-20
fix(plugin): correct enum+nullable handling for `Enum | null` (and ar…

AliRaZa1121:
…rays); avoid lazy `type` for enum unions

Context:
When a DTO property used a union like `Enum | null` (or `Enum[] | null`) the plugin sometimes:
- emitted a lazy `type: () => FooEnum` (unnecessary for enums; can cause circular refs), or
- missed `enum: FooEnum` entirely for optional+nullable cases (e.g. `?: Status | null`).

Root cause:
- `createTypePropertyAssignments` didn’t special-case enum unions, so it fell back to a lazy `type`.
- `createEnumPropertyAssignment` could return early when the checker reported a union containing null/undefined or when array unwrapping wasn’t detected up front, dropping the `enum`.

What changed:
1) Handle enum unions in `createTypePropertyAssignments`:
   - For unions with exactly one non-nullish member: * If that member (or its array element) is an enum (including auto-generated enum unions), do **not** emit `type`. * Only add `nullable: true` when `null` is part of the union.
   - This avoids circular resolvers and lets enum metadata be provided solely by `createEnumPropertyAssignment`.

2) Rewrite `createEnumPropertyAssignment` to be union- and optional-safe:
   - Unwrap `null`/\`undefined\` from unions before enum detection.
   - Treat array unwrapping as optional (don’t bail if not detected initially).
   - Resolve auto-generated enum unions (string literal unions) to a referenceable type.
   - Preserve and/or re-detect `isArray` after resolving enum-like types.
   - Skip enum **members** (e.g., `FooEnum.ONE`) which aren’t referenceable as enum types.

Results:
- `Enum | null` → `{ enum: FooEnum, nullable: true }`
- `Enum[] | null` → `{ enum: FooEnum, isArray: true, nullable: true }`
- `?: Enum | null` → `{ required: false, enum: FooEnum, nullable: true }`
- Non-enum unions (e.g., `string | null`) still emit `type` + `nullable` (as before).
- `undefined` continues to affect `required` (based on `?`), not `nullable` — unchanged.

Tests:
- Fixture `test/plugin/fixtures/nullable.dto.ts` asserts correct transpiled metadata.
- Previously failing spec “should support & understand nullable type unions” now passes.

Affects:
- lib/plugin/visitors/model-class.visitor.ts
- test/plugin/fixtures/nullable.dto.ts

Fixes: #3519

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/lib/plugin/visitors/model-class.visitor.ts
+++ b/lib/plugin/visitors/model-class.visitor.ts
@@ -675,7 +675,6 @@ export class ModelClassVisitor extends AbstractFileVisitor {
         type = nonNullish[0];
       }
     }
-    // <<<<<<<<< END ADD
 
     // Also handle cases where TypeScript emits an auto-generated union like `T | undefined`
     // (strict mode). In that case, pick the non-undefined member so enum detection works.


lib/plugin/visitors/model-class.visitor.ts
```diff
@@ -592,10 +649,44 @@ export class ModelClassVisitor extends AbstractFileVisitor {
...
+    if ((type.flags & ts.TypeFlags.Union) !== 0) {
+      const union = type as ts.UnionOrIntersectionType;
+      const nonNullish = union.types.filter(
+        (t) => (t.flags & (ts.TypeFlags.Null | ts.TypeFlags.Undefined)) === 0
+      );
+      if (nonNullish.length === 1) {
+        type = nonNullish[0];
+      }
+    }
+    // <<<<<<<<< END ADD
```
kamilmysliwiec:
```suggestion
```

PR #3603, opened 2025-10-23
fix(plugin): add async modifier when a reference is await import statement

seonggukchoi:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

When the CLI compiles comments and types into Swagger decorators, it omits the async keyword in arrow functions that reference another DTO, even though those functions use the await keyword.
This change ensures that any arrow function using await within a DTO reference property is correctly prefixed with the async keyword.

- [ ] Yes
- [x] No

## Other information

```json5
{
  // ...
  "compilerOptions": {
    "plugins": [{ "name": "@nestjs/swagger", "options": { "introspectComments": true, "esmCompatible": true } }]
  },
  // ...
}

```

```typescript
import { Exclude, Expose } from 'class-transformer';

@Exclude()
export class Foo {
  /**
   * Foo
   */
  @Expose()
  bar!: Bar;
}

@Exclude()
export class Bar {
  /**
   * Property
   */
  @Expose()
  property!: string;
}
```

```javascript
let Foo = class Foo {
        return { bar: { required: true, type: () => (await import("./foo-bar.dto.js")).Bar, description: "Foo" } }; // There is no 'async' keyword at the arrow function.
    }
};
```

### DIFF
--- a/lib/plugin/visitors/model-class.visitor.ts
+++ b/lib/plugin/visitors/model-class.visitor.ts
@@ -527,6 +527,10 @@ export class ModelClassVisitor extends AbstractFileVisitor {
       return [];
     }
 
+    const modifiers = typeReferenceDescriptor.typeName.includes('import(')
+      ? [factory.createModifier(ts.SyntaxKind.AsyncKeyword)]
+      : undefined;
+
     const identifier = typeReferenceToIdentifier(
       typeReferenceDescriptor,
       hostFilename,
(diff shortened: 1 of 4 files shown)


kamilmysliwiec:
This caused a regression https://github.com/nestjs/swagger/issues/3630

PR #3696, opened 2026-01-17
feat(schema): include class name in circular dependency error messages

### LINKED ISSUE #3655: Improve error message for circular dependencies in SchemaObjectFactory

When a circular dependency occurs during Swagger schema generation, the error message is generic: Error: A circular dependency has been detected (property key: "someKey").

In large projects or monorepos with hundreds of DTOs, knowing only the property name (e.g., "user", "filters", "id") is insufficient to locate the problematic class. It often requires tedious debugging or binary search (commenting out code) to find the source of the cycle.

I propose improving the error handling in SchemaObjectFactory, specifically within the extractPropertiesFromType method.

By wrapping the property extraction logic in a try-catch block, we can capture the error, prepend the name of the class (prototype) being processed, and re-throw it.

Proposed change logic:

```ts
// In SchemaObjectFactory.extractPropertiesFromType
try {
   // ... existing logic calling mergePropertyWithMetadata
} catch (err) {
   const className = type.name || 'Unknown';
   err.message = `[${className}] ${err.message}`;
   throw err;
}
```

This transforms the error from:
```ERROR MY.uncaughtException:Error: A circular dependency has been detected (property key: "someProperty") ...```

To:
```ERROR MY.uncaughtException:Error: [RootResponseDTO] [ParentEntity] [ChildEntity] A circular dependency has been detected (property key: "someProperty") ...```

This is a non-breaking change that significantly improves the Developer Experience (DX).

Adoption: Automatic. Users will simply see better error messages when they upgrade.
Documentation: No documentation changes required for the public API.
Migration: None needed.

The motivation is to drastically reduce debugging time. Circular dependencies are common in complex data models. Identifying the specific class causing the cycle allows developers to immediately apply the fix (using lazy resolvers type: () => Class or forwardRef) instead of hunting for the file.

### PR DESCRIPTION
**Pull request by ismaildasci**
## Summary

Improves the error message for circular dependencies by including the class name where the issue occurred. This makes debugging significantly easier in large projects with many DTOs.

## Problem

When a circular dependency is detected, the current error message only shows the property key:

```
Error: A circular dependency has been detected (property key: "user")
```

In large projects with hundreds of DTOs, knowing only the property name is insufficient to locate the problematic class.

## Solution

The error message now includes the class name(s) as errors bubble up:

```
Error: [UserResponseDto] [ParentEntity] A circular dependency has been detected (property key: "user")
```

## Implementation

- Added try-catch in `extractPropertiesFromType` to catch errors during property metadata extraction
- Prepends the class name to the error message before re-throwing
- Creates a path showing the full chain of classes involved in nested DTOs

## Benefits

1. **Faster Debugging**: Immediately identify which class has the circular dependency
2. **Non-Breaking**: Only changes error message format, no API changes
3. **Nested Path**: Shows full chain when DTOs reference each other

## Test Plan

- [x] Added test case for circular dependency error including class name

Closes #3655

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/lib/services/schema-object-factory.ts
+++ b/lib/services/schema-object-factory.ts
@@ -221,33 +221,41 @@ export class SchemaObjectFactory {
     const modelProperties =
       this.modelPropertiesAccessor.getModelProperties(prototype);
     const propertiesWithType = modelProperties.map((key) => {
-      const property = this.mergePropertyWithMetadata(
-        key,
-        prototype,
-        schemas,
-        pendingSchemasRefs
-      );
+      try {
+        const property = this.mergePropertyWithMetadata(
+          key,
+          prototype,
+          schemas,
+          pendingSchemasRefs
+        );
 
-      const schemaCombinators = ['oneOf', 'anyOf', 'allOf'];
-      const declaredSchemaCombinator = schemaCombinators.find(
-        (combinator) => combinator in property
-      );
-      if (declaredSchemaCombinator) {
-        const schemaObjectMetadata = property as SchemaObjectMetadata;
+        const schemaCombinators = ['oneOf', 'anyOf', 'allOf'];
+        const declaredSchemaCombinator = schemaCombinators.find(
+          (combinator) => combinator in property
+        );
+        if (declaredSchemaCombinator) {
+          const schemaObjectMetadata = property as SchemaObjectMetadata;
 
-        if (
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
In 99,9% cases the class name will be undefined (that's the problem with circular dependencies)

Issue #3274, opened 2025-01-26
Redocly lint error for nullable object properties


drewish:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

With the plugin enabled and looking at the Typescript code a property like:
```ts
  @ApiProperty({
    description: 'The user that owns the cat',
  })
  nullableOwner: UserDto | null;
```
, it generates the schema's property like:
```json
          "nullableOwner": {
            "nullable": true,
            "description": "The user that owns the cat",
            "allOf": [
              {
                "$ref": "#/components/schemas/UserDto"
              }
            ]
          }
```
[Redocly's linter](https://redocly.com/docs/cli/commands/lint) flags as an error:
```
[4] dist/api-spec.json:84:25 at #/components/schemas/CreateCatDto/properties/nullableOwner/nullable

The `type` field must be defined when the `nullable` field is used.

82 | },
83 | "nullableOwner": {
84 |   "nullable": true,
85 |   "description": "The user that owns the cat",
86 |   "allOf": [

Error was generated by the struct rule.
```

### Minimum reproduction code

https://github.com/drewish/swagger-nullable-poc

### Steps to reproduce

Setup the repro repo:
```
git clone https://github.com/drewish/swagger-nullable-poc
cd swagger-nullable-poc
npm i
```
Start it up to generate the swagger file (main.ts is saving a copy)
```
npm start
```
Then use redocly to lint the file:
```
npx @redocly/cli@latest lint dist/api-spec.json
```
There are unrelated errors I didn't bother cleaning up. The relevant one is:
```
The `type` field must be defined when the `nullable` field is used.
```

### Expected behavior

One workaround I've see for this (https://github.com/nestjs/swagger/issues/2948) was to just add `type: 'object'` as a sibling:
```
          "nullableOwner": {
            "type": "object",
            "nullable": true,
            "description": "The user that owns the cat",
            "allOf": [
              {
                "$ref": "#/components/schemas/UserDto"
              }
            ]
          }
```
It seems like in `ModelClassVisitor.createTypePropertyAssignments()`, when we create the nullable property:
https://github.com/nestjs/swagger/blob/8cb1ee8032c302a8a1cc226be053b90d4817bf52/lib/plugin/visitors/model-class.visitor.ts#L437-L443
We could also add `type: 'object'`. 

But, we'd also need a change in `SchemaObjectFactory.extractPropertiesFromType()` to keep it from deleting the type:
[...]

PR #3784, opened 2026-03-27
fix(schema): include type field when nullable is used with allOf

maruthang:
## PR Checklist
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added
- [x] Docs have been added / updated

## PR Type
- [x] Bugfix

When a property is `UserDto | null`, the schema has `nullable: true` and `allOf: [{ $ref }]` but no `type` field. Redocly's linter errors: "The type field must be defined when the nullable field is used."

Issue Number: #3274

## What is the new behavior?
Preserves `type: 'object'` for nullable properties with schema combinators. The `extractPropertiesFromType` combinator handler now keeps `type` when `nullable: true`.

- [x] No


kamilmysliwiec:
lgtm

PR #3803, opened 2026-04-02
fix(plugin): skip auto-generated response decorator when Api*Response already present

### LINKED ISSUE #1639: Adding @ApiFoundResponse on a dynamic redirect method still renders a 200 response code in the OpenAPI contract

### Current behavior

It appears that the Swagger CLI is adding an extra 200 response code to a method that will never return a 200 response. Specifically, for a method that always returns a dynamic redirect that's been labeled with `@ApiFoundResponse`

Example:
```typescript
  interface RedirectResponse {
    url: string
    statusCode?: number
  }

  @Get()
  @ApiFoundResponse({ description: 'Redirects to a URL' })
  @Redirect()
  getRedirect(): RedirectResponse {
    return {
      url: `https://example.com`,
      statusCode: HttpStatus.FOUND,
    }
  }
```

Which produces:
```json
  "/example": {
      "get": {
        ...
        "responses": {
          "200": {
            "description": "",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object"
                }
              }
            }
          },
          "302": {
            "description": "Redirects to a URL"
          }
        }
      }

```

### Minimum reproduction code

See above.

### Steps to reproduce

See above.

### Expected behavior

Only a single response is rendered - no 200 for a redirect.

### Package version

5.1.3

### NestJS version

8.0.11

### Node.js version

14

- [ ] macOS
- [ ] Windows
- [X] Linux

### Other

_No response_

maruthang:
## Summary

Fixes #1639

**Bug:** `@ApiFoundResponse` (and other `@Api*Response` decorators) were being ignored by the plugin, which always appended a default `ApiResponse` decorator with status 200/201, overriding the explicit response decorator.

**Root Cause:** `addDecoratorToNode` in `controller-class.visitor.ts` unconditionally appended an auto-generated `ApiResponse` decorator regardless of whether the method already had an explicit `@Api*Response` decorator.

**Fix:** Added a check — if any existing decorator name matches `ApiResponse` exactly or matches the `Api*Response` pattern, the auto-generated default is skipped entirely.

## Changes

- `lib/plugin/visitors/controller-class.visitor.ts`: Added `hasExplicitApiResponseDecorator` check before creating the auto-generated response decorator; only appends it when no explicit `@Api*Response` is present.
- `test/plugin/fixtures/app.controller.ts`: Updated existing snapshot fixture to reflect the corrected output.
- `test/plugin/fixtures/app.controller-api-response.ts`: New fixture with a method decorated with `@ApiFoundResponse` to serve as regression test input.
- `test/plugin/controller-class-visitor.spec.ts`: New regression test that verifies no extra `200` response is added when `@ApiFoundResponse` is already present.

## Testing

- Added regression test in `test/plugin/controller-class-visitor.spec.ts` that verifies no auto-generated 200 response is injected when `@ApiFoundResponse` is already present on a method.
- All existing tests pass (159 passing).


PeterTheOne:
This unfortunately seems to have caused a regression: #3862

kamilmysliwiec:
cc @maruthang

Issue #3591, opened 2025-10-15
Cannot find package '@nestjs/swagger' when using "type": "module" in package.json


lusu007:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

When using `@nestjs/swagger` in a project with ESM enabled (`"type": "module"` in package.json), the application fails to start with the following error:

```
[1] Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@nestjs/swagger' imported from /home/.../dist/apps/server/main.js
[1] Did you mean to import "@nestjs/swagger/dist/index.js"?
[1]     at Object.getPackageJSONURL (node:internal/modules/package_json_reader:316:9)
[1]     at packageResolve (node:internal/modules/esm/resolve:768:81)
[1]     at moduleResolve (node:internal/modules/esm/resolve:858:18)
[1]     at defaultResolve (node:internal/modules/esm/resolve:990:11)
[1]     at #cachedDefaultResolve (node:internal/modules/esm/loader:757:20)
[1]     at ModuleLoader.resolve (node:internal/modules/esm/loader:734:38)
[1]     at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:317:38)
[1]     at #link (node:internal/modules/esm/module_job:208:49) {
[1]   code: 'ERR_MODULE_NOT_FOUND'
[1] }
[1] 
[1] Node.js v24.10.0
[0] node:internal/modules/run_main:107
[0]     triggerUncaughtException(
[0]     ^
[0] 
[0] Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@nestjs/swagger' imported from /home/.../dist/apps/player/main.js
[0] Did you mean to import "@nestjs/swagger/dist/index.js"?
[0]     at Object.getPackageJSONURL (node:internal/modules/package_json_reader:316:9)
[0]     at packageResolve (node:internal/modules/esm/resolve:768:81)
[0]     at moduleResolve (node:internal/modules/esm/resolve:858:18)
[0]     at defaultResolve (node:internal/modules/esm/resolve:990:11)
[0]     at #cachedDefaultResolve (node:internal/modules/esm/loader:757:20)
[0]     at ModuleLoader.resolve (node:internal/modules/esm/loader:734:38)
[0]     at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:317:38)
[0]     at #link (node:internal/modules/esm/module_job:208:49) {
[...]

PR #3818, opened 2026-04-10
fix(core): add exports field to package.json for ESM compatibility

maruthang:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When a project uses `"type": "module"` in its `package.json`, importing `@nestjs/swagger` fails because Node.js ESM resolution requires an `exports` field to resolve bare specifiers. Without it, consumers get errors like `Cannot find package '@nestjs/swagger'` or `ERR_PACKAGE_PATH_NOT_EXPORTED`.

Issue Number: #3591

## What is the new behavior?

The `package.json` now includes an `exports` field with entries for:
- `"."` — main entry point (`dist/index.js` + types)
- `"./plugin"` — plugin subpath (`dist/plugin/index.js` + types)
- `"./package.json"` — allows reading the package metadata

This enables proper ESM resolution while maintaining full backward compatibility with CommonJS consumers via the existing `main` and `types` fields.

- [ ] Yes
- [x] No

## Other information

6 regression tests added in `test/esm-exports.spec.ts` to verify the exports field structure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,17 @@
   "repository": "https://github.com/nestjs/swagger",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./plugin": {
+      "types": "./dist/plugin/index.d.ts",
+      "default": "./dist/plugin/index.js"
+    },
+    "./package.json": "./package.json"
+  },
   "scripts": {
     "build": "tsc -p tsconfig.build.json",
     "format": "prettier \"lib/**/*.ts\" --write",
--- a/test/esm-exports.spec.ts
+++ b/test/esm-exports.spec.ts
@@ -1,0 +1,41 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+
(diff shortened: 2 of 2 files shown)


kamilmysliwiec:
This isnt sufficient to make this package esm compatible

maruthang:
Fair point @kamilmysliwiec — you're right that just adding the `exports` field isn't sufficient for real ESM compatibility. True dual-package support would also require:

1. Dual transpilation (CJS + ESM) with separate output directories, so `exports.import` points at actual ESM files rather than the existing CJS output
2. Updating the `main`/`module` fields to align with the new `exports` map
3. Handling the internal `.js` imports so they resolve under both module systems
4. Build pipeline changes (likely a separate tsc pass with `module: ESNext`, or a bundler step)

That's a bigger lift than this PR's scope. Would you prefer I:
- **Close this PR** and open an issue to track full ESM migration as a dedicated effort, or  
- **Expand this PR** to do the full dual-emit + exports mapping?

Happy to do either — just let me know which path you'd like!

kamilmysliwiec:
We can just migrate to ESM as that's the plan for other packages for v12

PR #3847, opened 2026-04-14
fix(plugin): adjust relative import paths for tsconfig outDir depth

### LINKED ISSUE #2706: Wrong path resolution in monorepo

### Current behavior

This issue is almost a duplicate of https://github.com/nestjs/swagger/issues/2114, I can't figure out why this should be address in Discord since it is a real bug and we are facing exactly the same issue in a pnpm workspace.

I can get it working by performing this manual change on dist:
- Open `apps/nest/dist/src/app.controller.js`
- Change `require("../../../packages/math-helpers/dist/index").Me` to `require("../../../../packages/math-helpers/dist/index").Me` 

Possible related issues: 
https://github.com/nestjs/swagger/issues/1386
https://github.com/nestjs/swagger/issues/501

### Minimum reproduction code

https://github.com/bindermuehle/monorepo-issue

### Steps to reproduce

1. `npm i`
2. `npm run start:dev -w nest`

### Expected behavior

nest/swagger should correctly resolve the paths

### Package version

7.1.10

### NestJS version

9.0.0

### Node.js version

20.9.0

- [X] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

kyungseopk1m:
## Summary

Fixes incorrect relative `require()`/`import()` paths generated by the CLI plugin in monorepo setups where `tsconfig` `outDir` is deeper than `rootDir`.

Closes #2706

## Problem

The plugin computes import paths using `posix.relative(sourceDir, importTarget)`. However, the generated code runs from the **output** file, not the source file. When `outDir` adds extra directory depth (e.g. source is `apps/nest/src/` but output lands in `apps/nest/dist/src/`), the path is off by the depth difference.

**Example**

| | Path |
|---|---|
| `rootDir` | `apps/nest` |
| `outDir` | `apps/nest/dist` |
| Source file | `apps/nest/src/app.controller.ts` |
| Output file | `apps/nest/dist/src/app.controller.js` |
| Import target | `packages/math-helpers/dist/index` |
| Generated (wrong) | `../../../packages/math-helpers/dist/index` |
| Correct | `../../../../packages/math-helpers/dist/index` |

## Fix

Introduce `getOutputDir()` in `plugin-utils.ts` that rebases the source directory onto the output tree before calling `posix.relative()`:

```
effectiveFrom = outDir + posix.relative(rootDir, sourceDir)
```

`outDir` and `rootDir` are read from `program.getCompilerOptions()` in `compiler-plugin.ts` and threaded through `PluginOptions` as `@internal` fields (not part of the public API). When they are absent the behaviour is unchanged, so this is fully backwards compatible.

## Changes

- `lib/plugin/merge-options.ts` — add `@internal` `outDir?` and `rootDir?` to `PluginOptions`
- `lib/plugin/compiler-plugin.ts` — populate `outDir`/`rootDir` from `program.getCompilerOptions()` after `mergePluginOptions()`
- `lib/plugin/utils/plugin-utils.ts` — add `getOutputDir()` helper; use it in `replaceImportPath()` instead of bare `posix.dirname(fileName)`
- `test/plugin/plugin-utils.spec.ts` — add 3 unit tests covering the bug scenario, the no-op case, and the absent-options fallback

## Testing

All 187 existing tests pass. Three new unit tests added:

- `outDir` deeper than `rootDir` → correct 4-level-up path (the bug scenario)
- `outDir === rootDir` → same result as without the fix (no-op)
- No `outDir`/`rootDir` in options → original behaviour preserved


kamilmysliwiec:
lgtm

kamilmysliwiec:
@kyungseopk1m sounds like your PR might have caused a regression https://github.com/nestjs/swagger/issues/3857

kyungseopk1m:
Thanks for flagging this, and thanks @alex-all3dp for picking it up so quickly in #3858.

In hindsight the gap in #3847 is clear: I rebased only the `from` side via `getOutputDir` and didn't apply the mirror transformation to the import target. Pre-#3847 both sides lived in source coordinates, so `posix.relative()` produced correct paths by accident — moving just `from` to output coordinates made the asymmetry observable.

A test covering "import target within `rootDir`" would've caught this; I should have added one alongside the monorepo case. Appreciate the catch.

Issue #3154, opened 2024-11-07
OpenAPI metadata required relative paths are broken since v8


bladx:
### Did you read the migration guide?

- [X] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Versions

7.4.2 -> 8.0.0

### Describe the regression

In the dist folder, the OpenAPI metadata required relative paths are broken since the v8.
I have a DTO with a property of type enum (file declared in another folder).
With the 7.4.2 I had: `enum: require('../../other-module/some-enum')`, but with the 8.0.0 or later I now get: `enum: require('./src/modules/other-module/some-enum')`.
This is breaking my built app as the path is wrong.
Having the `"preserveSymlinks": true` tsconfig option or not does not change anything.

For now, I stay on v7.4.2 for this reason.
But I feel like this is a regression that should be fixed.

### Expected behavior

I would expect the same behavior as before so having a relative path (starting with `../..`) instead of the new one (starting with `./src`).

kamilmysliwiec:
Does this happen with v8.0.0 (or v8.0.2 etc)?

Could you provide a minimum reproduction repository?

bladx:
I tried with at least v8.0.0, v8.0.2 and v8.0.3 and got the same error as described.
Regarding the minimum reproduction repository, I'll try to do it as soon as I find time but currently I don't really have time sorry.

kamilmysliwiec:
We unfortunately can't help without a reproduction repository

bladx:
I started, but I cannot find what produce the following in my app DTO: `static _OPENAPI_METADATA_FACTORY() {`.
In my minimum reproduction repository, I have an empty DTO class (so no issue), but in my real app I do have this metadata factory with the new regression.
Would you know what causes the generation of this particular function?

kamilmysliwiec:
Could you test with v8.0.4?

bladx:
It actually does fix the issue!
Thanks a lot for the reactivity and the precious help on that, really appreciate.

gwen1230:
@kamilmysliwiec I have same issue in a monorepo when the imported type is in another module.

norbornen:
@kamilmysliwiec I have same issue in a monorepo when the imported type is in another module.

Expected:
```
const foo_entities_1 = require("@foo/entities");
...
openapi.ApiResponse({ status: 201, type: require("@foo/entities").Item }),
or
[...]

PR #2707, opened 2023-11-14
fix: adapt imports replacement with tsconfig outDir depth

mgrisole:
Closes https://github.com/nestjs/swagger/issues/2706

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
[
](https://github.com/nestjs/swagger/issues/2706)

Issue Number: 2706

## What is the new behavior?

This PR introduces a function `adaptPathWithConfig` which translate the `outDir` attribute from the user tsconfig.json config into a relative path prepending the previously resolved one from `replaceImportPath`.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

Issue #1386, opened 2021-06-09
Wrong relative require path for custom rootDir


Kolobok12309:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

I have `dto` with enum field, `@root/types` example package from this monorepo
```
import { Role } from '@root/types';
export class InputDto {
  id: number;
  role?: Role;
}
```

Result dist
```
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.InputDto = void 0;
const openapi = require("@nestjs/swagger");
class InputDto {
    static _OPENAPI_METADATA_FACTORY() {
        return { id: { required: true, type: () => Number }, role: { required: false, enum: require("../../types/dist/users/role.interface").Role } };
    }
}
exports.InputDto = InputDto;
```

## Expected behavior

Expected require path `../../../types/dist/users/role.interface` or `@root/types/...`

## Minimal reproduction of the problem with instructions
[Repo](https://github.com/Kolobok12309/nestjs-swagger-repro)
To reproduce you can change `rootDir` to `.` or mb `src/foo`

## Environment

<pre><code>
Nest version: 7.6.15

 
For Tooling issues:
- Node version: v12.14.0
- Platform:  Win 10

Others:

My `rootDir` is `.`
</code></pre>

Kolobok12309:
Is duplicate of #846

kamilmysliwiec:
Would you like to create a PR for this issue?

Kolobok12309:
Don't think i can do it without more bug

Kolobok12309:
It fix my problem, i use `yarn workspaces` which create symlinks in `node_modules` and `typescript` resolve it, mb it will help someone
```
// tsconfig.json
...
preserveSymlinks: true
...
```
Core of trouble it `symlinks`

PR #3858, opened 2026-04-20
fix(plugin): remap import paths within rootDir to outDir

alex-all3dp:
## Summary

When `rootDir` and `outDir` are set in tsconfig (required by TypeScript 6), the CLI plugin's `replaceImportPath` generates broken `require()` calls that navigate from `dist/` back into `src/`.

**Example:** A DTO at `src/common/dto/order.dto.ts` importing an enum from `src/types/currency.ts` produces:
```js
// dist/common/dto/order.dto.js
require("../../../src/types/currency")  // broken — src/ doesn't exist at runtime
```
instead of:
```js
require("../../types/currency")  // correct — resolves within dist/
```

This breaks in Docker containers and production deployments where only `dist/` is available.

## Root cause

`replaceImportPath` correctly rebases the *current* file's directory onto the output tree via `getOutputDir`, but uses the *source* path of the imported module as-is. `posix.relative(outputDir, sourceImportPath)` then produces a path that escapes `dist/` and navigates into `src/`.

## Fix

Apply the same `rootDir` → `outDir` transformation to the import target path when it falls within `rootDir`, so both sides of `posix.relative()` are in the output coordinate space. Paths outside `rootDir` (e.g. monorepo cross-package imports) are left unchanged.

## Context

- TypeScript 6 makes `rootDir` mandatory (TS5011 error) when the common source directory differs from the tsconfig location
- This regression was introduced alongside the `getOutputDir` fix in #2707 — the current-file side was corrected but the import-target side was missed
- Related: #3154, #1386

## Test plan

- [x] Added test case for `rootDir = src/` with import within rootDir
- [x] Existing monorepo test (import outside rootDir) still passes
- [x] All 247 tests pass
- [x] Verified fix against real NestJS project with `@nestjs/swagger` plugin + TypeScript 6

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/lib/plugin/utils/plugin-utils.ts
+++ b/lib/plugin/utils/plugin-utils.ts
@@ -197,7 +197,15 @@ export function replaceImportPath(
       ? safeDecodeURIComponent(convertPath(options.pathToSource))
       : getOutputDir(fileName, options);
 
-    let relativePath = posix.relative(from, decodedImportPath);
+    const targetPath =
+      options.outDir && options.rootDir &&
+      decodedImportPath.startsWith(convertPath(options.rootDir))
+        ? posix.join(
+            convertPath(options.outDir),
+            posix.relative(convertPath(options.rootDir), decodedImportPath)
+          )
+        : decodedImportPath;
+    let relativePath = posix.relative(from, targetPath);
     relativePath = relativePath[0] !== '.' ? './' + relativePath : relativePath;
 
     const normalizedPath = normalizePackagePath(relativePath);
--- a/test/plugin/plugin-utils.spec.ts
+++ b/test/plugin/plugin-utils.spec.ts
@@ -150,6 +150,31 @@ describe('plugin-utils', () => {
       expect(result.typeReference).toContain('../entities/test.entity');
     });
 
+    it('should remap import paths within rootDir to outDir (rootDir = src/)', () => {
+      // Standard layout with rootDir pointing to src/:
+      //   rootDir : /project/src
+      //   outDir  : /project/dist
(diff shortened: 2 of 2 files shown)


kamilmysliwiec:
sounds like this aims to fix the regression introduced by https://github.com/nestjs/swagger/pull/3847?

alex-all3dp:
Not a regression from #3847 — that PR correctly fixed the `from` side (rebasing the current file's directory onto the output tree via `getOutputDir`). This PR fixes the `to` side: the imported module's source path also needs the same rootDir→outDir transformation when it falls within rootDir. Without this, `posix.relative(outputDir, sourceImportPath)` produces a path that escapes `dist/` back into `src/`.

The issue predates #3847 but was masked because without `getOutputDir`, both `from` and `to` were in source coordinates, so `posix.relative()` produced correct relative paths between source files. With #3847, `from` moved to output coordinates while `to` stayed in source coordinates — making the mismatch visible.

kyungseopk1m:
Thanks for the quick fix!

alex-all3dp:
Correction to my previous comment — you're right, this is a regression from #3847. On 11.3.0 (without `getOutputDir`), both sides of `posix.relative()` were in source coordinates, so paths were correct. #3847 moved the `from` side to output coordinates via `getOutputDir` but left the `to` side (the imported module path) in source coordinates, creating the mismatch. This PR applies the same transformation to the `to` side.

alex-all3dp:
Thanks for the quick release @kamilmysliwiec !

PR #3863, opened 2026-04-21
fix(plugin): restore default 2xx response when only error Api*Response decorators present

### LINKED ISSUE #3862: Regression in 11.2.7: plugin drops auto-inferred 2xx response when only error @Api*Response decorators are present (PR #3803)

- [x] I have read the whole migration guide

PR #3803 (released in 11.2.7)

### Versions

11.2.6 → 11.2.7 (regression still present in 11.3.2)

### Describe the regression

PR #3803 (fixes #1639) added a `hasExplicitApiResponseDecorator` check that suppresses the auto-inferred default response when any decorator whose name matches `Api*Response` is present. The check doesn't distinguish success (2xx) from error (4xx/5xx) status codes — so a handler decorated with only `@ApiUnauthorizedResponse()` (or `@ApiNotFoundResponse()`, `@ApiForbiddenResponse()`, etc.) no longer gets the auto-inferred 201/200 response in the OpenAPI spec.

The resulting spec lists only the error status code. Downstream client generators (NSwag, openapi-generator, etc.) can't tell that a 2xx is legitimate and treat the real 2xx response as an exception at runtime.

Before 11.2.7, the default 201/200 was always appended, so these handlers produced a complete spec even without an explicit success decorator.

### Minimum reproduction code

```ts
import { Controller, Post } from '@nestjs/common';
import { ApiUnauthorizedResponse } from '@nestjs/swagger';

@Controller('session')
export class SessionController {
  @Post('start')
  @ApiUnauthorizedResponse()
  startSession(): void {
    // defaults to 201 Created at runtime
  }
}
```

Resulting OpenAPI spec:

```yaml
# 11.2.6 (correct)
/session/start:
  post:
    responses:
      '201': { description: '' }       # ← auto-inferred, correct
      '401': { description: '' }

# 11.2.7+ (broken)
/session/start:
  post:
    responses:
      '401': { description: '' }       # ← 201 silently dropped
```

### Expected behavior

The auto-inferred 2xx response should only be suppressed when at least one *success* `@Api*Response` decorator (2xx) is already declared on the handler. Error-only decorators (`@ApiUnauthorizedResponse`, `@ApiNotFoundResponse`, `@ApiForbiddenResponse`, `@ApiBadRequestResponse`, `@ApiConflictResponse`, etc.) should not suppress the default.

### Other

Real-world impact: in our codebase, a `POST /portal-session/start` handler with just `@ApiUnauthorizedResponse()` started throwing `ApiException` on every successful login after we bumped to 11.2.7, because the generated Angular client had no entry for 201 in its response handler. Downgrading to 11.2.6 confirmed the regression.

yogeshwaran-c:

Bug fix.

A regression introduced in 11.2.7 (by #3803, fixing #1639) causes the plugin to drop the auto-inferred 2xx response whenever a controller method is decorated with only error `@Api*Response` decorators. The `hasExplicitApiResponseDecorator` guard added in that PR treats **any** decorator whose name matches `Api*Response` as an explicit response, so handlers annotated with only `@ApiUnauthorizedResponse()`, `@ApiNotFoundResponse()`, `@ApiResponse({ status: 500 })`, etc. lose their auto-inferred 201/200. The resulting spec advertises only the error status codes, which breaks downstream client generators (NSwag, openapi-generator, etc.).

Before 11.2.7:

```yaml
/session/start:
  post:
    responses:
      '201': { description: '' }
      '401': { description: '' }
```

After 11.2.7 (broken):

```yaml
/session/start:
  post:
    responses:
      '401': { description: '' }
```

Closes #3862

## What is the new behavior?

The plugin now treats a decorator as an explicit success declaration only when:

- it is a 2xx-coded `Api*Response` alias (e.g. `ApiOkResponse`, `ApiCreatedResponse`, `ApiNoContentResponse`, `ApiAcceptedResponse`, and the other 2xx aliases derived from `HttpStatus`),
- it is `ApiDefaultResponse`, or
- it is a plain `@ApiResponse` whose `status` argument is a 2xx numeric literal, `'2XX'`, or `'default'`.

Non-success decorators (3xx/4xx/5xx, including `@ApiFoundResponse`, `@ApiUnauthorizedResponse`, and friends) no longer suppress the inferred 2xx. The auto-generated default is emitted alongside the explicit error entries, and the OpenAPI spec once again contains both success and error status codes.

## Additional context

- The set of 2xx alias names is derived from `HttpStatus` at module-load time, mirroring how the decorators themselves are generated in `lib/decorators/api-response.decorator.ts`, so any future additions stay in sync automatically.
- When the `status` argument of `@ApiResponse` is a non-literal expression (e.g. `HttpStatus.OK`), the plugin cannot evaluate it at compile time; to preserve the original intent of #1639, such expressions continue to be treated as explicit success declarations.
- A dedicated regression fixture (`test/plugin/fixtures/app.controller-error-only-api-response.ts`) exercises the error-only scenarios from the issue: a POST with `@ApiUnauthorizedResponse`, a GET with multiple error decorators, and a `@ApiResponse({ status: 500 })`. The existing `#1639` fixture is updated so the 302 redirect handler now carries the restored default alongside its explicit `@ApiFoundResponse`.
- `npm test` (274 unit tests) and `npm run test:e2e` (114 tests) pass locally.

### DIFF
--- a/lib/plugin/visitors/controller-class.visitor.ts
+++ b/lib/plugin/visitors/controller-class.visitor.ts
@@ -1,3 +1,4 @@
+import { HttpStatus } from '@nestjs/common';
 import { compact, head } from 'lodash';
 import { posix } from 'path';
 import * as ts from 'typescript';
@@ -24,6 +25,57 @@ import { typeReferenceToIdentifier } from '../utils/type-reference-to-identifier
 import { AbstractFileVisitor } from './abstract.visitor';
 
 type ClassMetadata = Record<string, ts.ObjectLiteralExpression>;
+
+const SUCCESS_API_RESPONSE_DECORATORS = new Set(
+  Object.keys(HttpStatus)
+    .filter((key) => {
+      const code = Number(HttpStatus[key as keyof typeof HttpStatus]);
+      return !isNaN(code) && code >= 200 && code < 300;
+    })
+    .map((key) => {
+      const functionName = key
+        .split('_')
+        .map(
+          (strToken) =>
+            `${strToken[0].toUpperCase()}${strToken.slice(1).toLowerCase()}`
+        )
+        .join('');
+      return `Api${functionName}Response`;
+    })
+    .concat(['ApiDefaultResponse'])
+);
+
+function isSuccessStatusArgument(decorator: ts.Decorator): boolean {
+  const args = getDecoratorArguments(decorator);
+  const firstArg = head(args);
+  if (!firstArg || !ts.isObjectLiteralExpression(firstArg)) {
+    return false;
+  }
+  const statusProp = firstArg.properties.find(
+    (prop) =>
+      ts.isPropertyAssignment(prop) &&
+      ((ts.isIdentifier(prop.name) && prop.name.text === 'status') ||
+        (ts.isStringLiteral(prop.name) && prop.name.text === 'status'))
+  ) as ts.PropertyAssignment | undefined;
+  if (!statusProp) {
+    return false;
+  }
+  const initializer = statusProp.initializer;
+  if (ts.isNumericLiteral(initializer)) {
+    const code = Number(initializer.text);
+    return code >= 200 && code < 300;
+  }
+  if (ts.isStringLiteral(initializer)) {
+    const value = initializer.text;
+    return value === '2XX' || value === 'default';
+  }
+  // Fallback for property access expressions like HttpStatus.OK or other
+  // identifiers that resolve to a 2xx value at runtime. We cannot evaluate
+  // the expression here, so we default to treating it as an explicit
+  // response declaration (preserves the original behavior of the
+  // explicit-decorator guard).
+  return true;
+}
 
 export class ControllerClassVisitor extends AbstractFileVisitor {
   private readonly _collectedMetadata: Record<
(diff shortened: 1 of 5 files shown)


PeterTheOne:
Heads up — this re-introduces #1639. The 2xx-only filter excludes `ApiFoundResponse` (302), so `@Get @Redirect @ApiFoundResponse` gets a spurious 200 added back. The fixture diff in this PR asserts exactly that:

```diff
     (0, swagger_1.ApiFoundResponse)({ description: 'Redirects to a URL' }),
-    (0, common_1.Redirect)()
+    (0, common_1.Redirect)(),
+    openapi.ApiResponse({ status: 200, type: Object })
```

yogeshwaran-c:
Thanks @PeterTheOne — good catch, the fixture diff was indeed showing the bug, not the fix. Pushed f00a9fc0.

The underlying issue was that the "explicit declaration" guard only recognised 2xx decorators. For a handler with `@Get + @Redirect + @ApiFoundResponse (302)`, none of those were 2xx-coded, so the plugin fell back to the auto-inferred 200 — reproducing #1639 for the 3xx case.

**Fix:** widen the non-error decorator set (renamed `SUCCESS_API_RESPONSE_DECORATORS` → `NON_ERROR_API_RESPONSE_DECORATORS`) and the `@ApiResponse({ status })` check to cover 2xx **and** 3xx. Redirect-style handlers now suppress the auto 200; error-only decorators (4xx/5xx) still allow the default to be emitted, preserving the #3862 behaviour.

The `#1639` fixture now asserts the corrected output — no auto-generated `ApiResponse({ status: 200, type: Object })` on the redirect handler — and I added a `@ApiResponse({ status: 301 })` case to exercise the numeric-literal branch for 3xx too.

kamilmysliwiec:
fixed https://github.com/nestjs/swagger/pull/3867

yogeshwaran-c:
Thanks @kamilmysliwiec — happy this is fixed via #3867. Closing the loop on this one.

PR #3865, opened 2026-04-21
feat(schema-object-factory): include class name chain in circular dependency errors

### LINKED ISSUE #3655: Improve error message for circular dependencies in SchemaObjectFactory

yogeshwaran-c:

Feature / DX improvement in `SchemaObjectFactory`.

When a circular dependency is detected during Swagger schema generation, the error message only identifies the property key involved:

```
Error: A circular dependency has been detected (property key: "someKey"). Please, make sure that each side of a bidirectional relationships are using lazy resolvers ("type: () => ClassType").
```

In a large codebase (e.g. a monorepo with hundreds of DTOs), knowing only the property name is often not enough to locate the offending class — developers end up commenting code out or running a binary search to identify the cycle.

Closes #3655

## What is the new behavior?

`extractPropertiesFromType` now wraps the per-property extraction in a `try/catch` that prepends the currently-processed class name to the error message and re-throws. As the recursion unwinds, the error accumulates the full chain of classes that led to the failure:

```
Error: [RootResponseDto] [ParentEntity] [ChildEntity] A circular dependency has been detected (property key: "someProperty"). To resolve this, use a lazy resolver for the property type ("type: () => ClassType") on each side of the relationship, or break the cycle by introducing a reference via @ApiExtraModels.
```

The base error hint was also extended to mention `@ApiExtraModels` alongside lazy resolvers, since either approach can break a cycle.

This is a non-breaking change: existing error handling that already catches this error will continue to work, and the only user-visible difference is a more informative message.

## Additional context

- The implementation follows the approach suggested in the issue by the original reporter.
- A regression test was added in `test/services/schema-object-factory.spec.ts` that simulates a circular dependency across two DTOs and asserts that both class names appear in the thrown error message.
- The guard `if (!err.message.startsWith(prefix))` avoids double-prefixing when the same class happens to appear twice consecutively in the chain (defensive — unlikely in practice).

### DIFF
--- a/lib/services/schema-object-factory.ts
+++ b/lib/services/schema-object-factory.ts
@@ -221,12 +221,29 @@ export class SchemaObjectFactory {
     const modelProperties =
       this.modelPropertiesAccessor.getModelProperties(prototype);
     const propertiesWithType = modelProperties.map((key) => {
-      const property = this.mergePropertyWithMetadata(
-        key,
-        prototype,
-        schemas,
-        pendingSchemasRefs
-      );
+      let property: ReturnType<typeof this.mergePropertyWithMetadata>;
+      try {
+        property = this.mergePropertyWithMetadata(
+          key,
+          prototype,
+          schemas,
+          pendingSchemasRefs
+        );
+      } catch (err) {
+        // Prepend the current class name to the error message so that, as the
+        // recursion unwinds, users see the full chain of classes involved in
+        // the failure (e.g. "[RootDto] [ParentDto] [ChildDto] A circular
+        // dependency has been detected ..."). This makes it much easier to
+        // locate the offending class in large codebases.
+        if (err instanceof Error) {
+          const className = type?.name || 'UnknownType';
+          const prefix = `[${className}] `;
+          if (!err.message.startsWith(prefix)) {
+            err.message = `${prefix}${err.message}`;
+          }
+        }
+        throw err;
+      }
 
       const schemaCombinators = ['oneOf', 'anyOf', 'allOf'];
       const declaredSchemaCombinator = schemaCombinators.find(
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #3867, opened 2026-04-21
fix(plugin): keep auto-inferred default response when only error Api*Response decorators are present

### LINKED ISSUE #3862: Regression in 11.2.7: plugin drops auto-inferred 2xx response when only error @Api*Response decorators are present (PR #3803)

PeterTheOne:
## Summary

Fixes #3862 — regression introduced in 11.2.7 via PR #3803.

PR #3803 added a guard that suppresses the auto-inferred default response whenever *any* `Api*Response` decorator is present. The guard doesn't distinguish success (2xx) from error (4xx/5xx) decorators, so a handler with only `@ApiUnauthorizedResponse()` (or any other error-status decorator) loses its auto-inferred 2xx in the OpenAPI spec. Downstream client generators (NSwag, openapi-generator, ...) then treat the real 2xx response as an exception at runtime.

This PR scopes the guard to decorators whose status code is **below 400** so error factories no longer suppress the default. The redirect case the original PR fixed (`@ApiFoundResponse` on a `@Redirect()` handler) continues to work — 302 is still below 400.

## Approach

**Named `Api*Response` factories** — status codes are derived from the decorator name via the same `Api${PascalCase(HttpStatus key)}Response` convention the factories are generated from in `api-response.decorator.ts`. Unknown names (`ApiDefaultResponse`, user-defined `Api*Response`) fall back to the previous behavior (treated as explicit), so no existing explicit-response pattern regresses.

Mirrors the forward-direction idiom in `api-response.decorator.ts:148`:
```ts
// forward: generating factories
.filter((key) => !isNaN(Number(HttpStatus[key])))
// reverse: decoder in this PR
const status = Number(HttpStatus[statusKey as keyof typeof HttpStatus]);
return isNaN(status) || status < 400;
```

**Plain `@ApiResponse({ status })`** — the `status` argument is inspected: numeric literals (`status: 500`), the shorthand strings `'1XX'`/`'2XX'`/`'3XX'`, and `'default'` are classified by their code. Non-literal expressions (e.g. `HttpStatus.OK`) can't be evaluated at compile time and keep the pre-PR behavior (treated as explicit).

## Behavior matrix

| Handler decorators | Before (11.2.7–11.3.2) | After this PR |
|---|---|---|
| `@Post` + `@ApiUnauthorizedResponse` | only 401 ❌ | 201 + 401 ✅ |
| `@Get` + `@ApiNotFoundResponse` | only 404 ❌ | 200 + 404 ✅ |
| `@Post` + `@ApiResponse({ status: 500 })` | only 500 ❌ | 201 + 500 ✅ |
| `@Post` + `@ApiCreatedResponse` + `@ApiUnauthorizedResponse` | 201 + 401 ✅ | 201 + 401 ✅ (unchanged) |
| `@Get @Redirect @ApiFoundResponse` (#1639) | 302 only ✅ | 302 only ✅ (unchanged) |

## Tests

- New fixture `test/plugin/fixtures/app.controller-api-response-error.ts` covering five scenarios:
  - `@Post` + `@ApiUnauthorizedResponse` alone
  - `@Get` + `@ApiNotFoundResponse` alone
  - `@Post` + `@ApiCreatedResponse` + `@ApiUnauthorizedResponse` (explicit success, default still suppressed)
  - `@Post` + `@ApiResponse({ status: 500 })` alone (arg-inspection case)
  - `@Get` + `@Redirect` + `@ApiFoundResponse` (#1639 preservation)
- New regression test in `controller-class-visitor.spec.ts` (named after #3862).
- Existing #1639 fixture (`app.controller-api-response.ts`) still passes — redirect case preserved.
- Existing `app.controller` fixture updated: its `create()` handler uses multiple error-status `@ApiResponse` decorators and the expected output had baked in the 11.2.7 regression. Now expects the restored pre-11.2.7 behavior (default 201 alongside the error entries).
- Full test suite: 274/274 passing. `npm run lint` clean on changed files. `prettier --check` clean on changed files.

## Disclaimers

- Transparency: this PR was prepared in collaboration with an AI coding assistant. I reviewed and approved every change. Happy to iterate on feedback.
- Given the subtlety of the 11.2.7 regression (silent OpenAPI spec diff, only surfaces in downstream client generators), reviewers may want to independently verify no other response-generation paths are affected.

### DIFF
--- a/lib/plugin/visitors/controller-class.visitor.ts
+++ b/lib/plugin/visitors/controller-class.visitor.ts
@@ -1,3 +1,4 @@
+import { HttpStatus } from '@nestjs/common';
 import { compact, head } from 'lodash';
 import { posix } from 'path';
 import * as ts from 'typescript';
@@ -170,11 +171,15 @@ export class ControllerClassVisitor extends AbstractFileVisitor {
     const hasExplicitApiResponseDecorator = decorators.some((item) => {
       try {
         const decoratorName = getDecoratorName(item);
-        return (
-          decoratorName === ApiResponse.name ||
-          (decoratorName.startsWith('Api') &&
-            decoratorName.endsWith('Response'))
-        );
+        if (decoratorName === ApiResponse.name) return true;
+        const statusNameMatch = decoratorName.match(/^Api(.+)Response$/);
+        if (!statusNameMatch) return false;
+        const statusKey = statusNameMatch[1]
+          .replace(/([a-z0-9])([A-Z])/g, '$1_$2')
+          .toUpperCase();
+        const status = Number(HttpStatus[statusKey as keyof typeof HttpStatus]);
+        // Error factories (4xx/5xx) must not suppress the auto-inferred 2xx.
+        return isNaN(status) || status < 400;
       } catch {
         return false;
       }
--- a/lib/plugin/visitors/controller-class.visitor.ts
+++ b/lib/plugin/visitors/controller-class.visitor.ts
@@ -26,6 +26,30 @@ import { AbstractFileVisitor } from './abstract.visitor';
 
 type ClassMetadata = Record<string, ts.ObjectLiteralExpression>;
 
+function isSuccessOrRedirectApiResponseArg(decorator: ts.Decorator): boolean {
+  const [firstArg] = getDecoratorArguments(decorator);
+  if (!firstArg || !ts.isObjectLiteralExpression(firstArg)) return true;
+  const statusProp = firstArg.properties.find(
+    (p): p is ts.PropertyAssignment =>
+      ts.isPropertyAssignment(p) &&
+      ts.isIdentifier(p.name) &&
+      p.name.text === 'status'
+  );
+  if (!statusProp) return true;
+  const init = statusProp.initializer;
+  if (ts.isNumericLiteral(init)) return Number(init.text) < 400;
+  if (ts.isStringLiteral(init)) {
+    return (
+      init.text === '1XX' ||
+      init.text === '2XX' ||
+      init.text === '3XX' ||
+      init.text === 'default'
+    );
+  }
+  // Non-literal (e.g. HttpStatus.OK) — can't evaluate at compile time; preserve pre-PR behavior.
+  return true;
+}
+
 export class ControllerClassVisitor extends AbstractFileVisitor {
   private readonly _collectedMetadata: Record<
     string,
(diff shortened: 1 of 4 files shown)


lib/plugin/visitors/controller-class.visitor.ts
```diff
@@ -25,6 +26,30 @@ import { AbstractFileVisitor } from './abstract.visitor';
 
 type ClassMetadata = Record<string, ts.ObjectLiteralExpression>;
 
+function isSuccessOrRedirectApiResponseArg(decorator: ts.Decorator): boolean {
```
kamilmysliwiec: nit: can we move it to another dedicated file, or at least make it a method of the ControllerClassVisitor class

kamilmysliwiec:
just checking to see if you're still willing to work on this PR @PeterTheOne, or if you would rather want us to focus on https://github.com/nestjs/swagger/pull/3863 instead

lib/plugin/visitors/controller-class.visitor.ts
PeterTheOne: Done — moved it to a private method on `ControllerClassVisitor` in fe80eb33. Felt like overkill to spin up a dedicated file for a ~20-line helper with one call site, but happy to extract further if you'd prefer.

PeterTheOne:
@kamilmysliwiec I did your nit comment. I don't mind if we go with the other PR, I just want this fixed asap :)

kamilmysliwiec:
excellent @PeterTheOne thank you! would you mind resolving PR conflicts? if so, ill get this merged and released asap

PeterTheOne:
Rebased onto current master, conflicts resolved (just a test-placement collision with the new `@Query` optional test). All plugin tests pass — should be ready to merge.

kamilmysliwiec:
LGTM

PR #3868, opened 2026-04-21
feat(plugin): auto-mark optional @Query parameters as required: false

### LINKED ISSUE #30: Automatically detect optional @Query
Am I right in thinking that you could look for a `?` on the query function parameter to denote that it is optional, and pick this up to set the `required` value on the swagger document? 

It would be nicer than having to add decorators like `@ApiImplicitQuery({name:'offset',required:false})` to the method I think. What do you reckon?

Thanks :)

yogeshwaran-c:

Feature.

When a controller method declares an optional query parameter like:

```ts
@Get()
findAll(@Query('limit') limit?: number) {}
```

the generated OpenAPI spec marks `limit` as required because the CLI plugin never infers optionality from the parameter signature. Users have been working around this by repeating the information in a manual `@ApiQuery({ name: 'limit', required: false })` decorator for years (long-standing request in #30).

## What is the new behavior?

The CLI plugin's controller visitor now inspects every method parameter. For each parameter that:

1. Is decorated with `@Query('name')` (a literal name — no behaviour change for `@Query()` whole-query-object form), and
2. Is optional at the TypeScript level:
   - `@Query('limit') limit?: number`
   - `@Query('offset') offset: number = 0`
   - `@Query('sort') sort: string | undefined`

it auto-emits `@ApiQuery({ name, required: false })` so the generated OpenAPI parameter is optional.

Safety guardrails:

- Existing `@ApiQuery({ name: 'foo', ... })` decorators on the method take precedence — no duplicates are generated for that name.
- If the method uses a shape of `@ApiQuery` we cannot reliably introspect at the AST level (non-object-literal argument, or missing literal `name`), generation is skipped for that whole method to avoid emitting conflicting metadata.
- Required parameters (no `?`, no default, no union with `undefined`) are left untouched — preserves the existing `required: true` default.
- Detection is AST-only so the transform keeps working in isolated `transpileModule` contexts where cross-file type information is not available.

Fixes #30

## Additional context

- New test: `should mark optional @Query parameters as required: false (issue #30)` in `test/plugin/controller-class-visitor.spec.ts`, with a dedicated fixture in `test/plugin/fixtures/app.controller-optional-query.ts` covering all three optional shapes, the "existing explicit `@ApiQuery` wins" case, and the bare-`@Query()` skip.
- All 274 existing tests continue to pass.


kamilmysliwiec:
great stuff, thanks

PR #3870, opened 2026-04-22
fix(decorators): forward all OpenAPI parameter fields in @ApiHeader

yogeshwaran-c:
Bug fix.

`@ApiHeader` only forwarded `description`, `required`, `examples` and `schema` onto the generated parameter metadata. The remaining OpenAPI Parameter Object fields it advertises via `ApiHeaderOptions extends Omit<ParameterObject, 'in'>` — `deprecated`, `allowEmptyValue`, `style`, `explode`, `allowReserved`, `content` — were silently dropped.

Minimal repro:

```ts
@ApiHeader({
  name: 'X-API-Version',
  deprecated: true,
  allowEmptyValue: true,
  style: 'simple',
  explode: false,
  allowReserved: true,
})
@Get()
find() {}
```

None of the flagged flags appear on the emitted parameter. A `content`-only header additionally ended up with an unwanted synthetic `schema: { type: 'string' }`, which is invalid per OpenAPI (`schema` and `content` are mutually exclusive on a Parameter Object).

## What is the new behavior?
- All fields defined on `BaseParameterObject` (`deprecated`, `allowEmptyValue`, `style`, `explode`, `allowReserved`, `content`) are now forwarded into the parameter metadata.
- When the caller supplies `content`, the default `schema: { type: 'string' }` is suppressed so the emitted parameter only carries `content`, matching the spec's mutual-exclusivity rule.
- Behaviour for existing callers is unchanged: the default `schema: { type: 'string' }` and the top-level `example` mapping to `schema.example` both still apply when no `content` is provided.

## Additional context
Added a dedicated `test/decorators/api-header.decorator.spec.ts` with four regressions (simple header, fully-populated header, `content`-only header, top-level `example`). The existing swagger-explorer `when headers are defined` suite continues to pass untouched.

### DIFF
--- a/lib/decorators/api-header.decorator.ts
+++ b/lib/decorators/api-header.decorator.ts
@@ -22,18 +22,31 @@ const defaultHeaderOptions: Partial<ApiHeaderOptions> = {
 export function ApiHeader(
   options: ApiHeaderOptions
 ): MethodDecorator & ClassDecorator {
+  // `schema` and `content` are mutually exclusive per the OpenAPI spec for a
+  // single parameter object, so only provide a default string schema when the
+  // caller has not opted in to `content`.
+  const paramSchema = options.content
+    ? undefined
+    : {
+        type: 'string',
+        ...(options.example ? { example: options.example } : {}),
+        ...(options.schema || {})
+      };
+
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
LGTM

PR #3873, opened 2026-04-22
fix(plugin): strip regex delimiters and flags from @Matches patterns

yogeshwaran-c:
Bug fix (plugin / class-validator bridge).

When a property is annotated with `@Matches(/^abc$/i)`, the plugin previously copied the whole decorator-argument text (`.text`) into the generated OpenAPI metadata. The resulting metadata is:

```js
pattern: { required: true, type: () => String, pattern: \"/^abc$/i\" }
```

The leading/trailing `/` and the `i` flag are not valid JSON-Schema / OpenAPI `pattern` syntax — tools like openapi-generator and the Swagger UI validator reject it. This is inconsistent with the runtime-only path, which goes through `RegExp.source` and therefore emits the bare pattern.

Proof: existing fixtures `create-cat.dto`, `create-cat-exclude.dto`, `create-cat-priority.dto`, and `serialized-meta*.fixture` all encode `pattern: \"/^[+]?abc$/\"`, i.e. they codify the buggy output today.

## What is the new behavior?
The Matches bridge now special-cases the two legal argument shapes for class-validator's `@Matches`:

- **RegularExpressionLiteral** — extract the source between the leading and trailing `/` and drop any flags. Matches what `new RegExp(value).source` returns at runtime. Flags are dropped because `pattern` in OpenAPI has no flag syntax (JSON Schema's ECMA-262 dialect does not support them).
- **StringLiteral** — use `.text` as-is; it already contains the bare pattern.

Anything else is ignored rather than guessed at.

After the fix, the same `@Matches(/^[+]?abc$/)` property emits:

```js
pattern: { required: true, type: () => String, pattern: \"^[+]?abc$\" }
```

## Additional context
- Updated all existing fixtures that previously encoded the buggy output (`create-cat.dto`, `create-cat-exclude.dto`, `create-cat-priority.dto`, `serialized-meta.fixture`, `serialized-meta-esm.fixture`). Each one was carrying forward the same leaked-slash pattern, which is the clearest possible demonstration of the bug.
- Added a new `create-cat-matches.dto` fixture plus a regression test in `model-class-visitor.spec.ts` that exercises three cases: `@Matches(/^abc$/i)` (RegExp with flags), `@Matches(/\d+/)` (RegExp with an escape sequence), and `@Matches('plain-string')` (StringLiteral fallback).


kamilmysliwiec:
LGTM

Issue #3335, opened 2025-03-06
@ApiProperty's example property being overridden


uPaymeiFixit:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

I have a `ParentDto` which has a property of type `ChildDto`. Both DTOs have `ApiProperty` decorators with `example` properties on them. I would expect the `example` from `ParentDto` to override any of the examples in `ChildDto`, which is the case when the DTO is used with `@Body`, but when it's used with `@Header`, `@Param`, or `@Query` the child DTO examples are used instead. 

@Papooch helped me find a workaround by adding `type: 'string'` to the `ParentDto`. But I'm still wondering, is this expected behavior? If so, why?

```typescript
import { Body, Controller, Get, Headers, Param, Query } from '@nestjs/common';
import { AppService } from './app.service';
import { ApiProperty } from '@nestjs/swagger';

class ChildDto {
  @ApiProperty({
    example: 'child DTO example 1',
  })
  childKey1: string;

  @ApiProperty({
    example: 'child DTO example 2',
  })
  childKey2: string;
}

class ParentDto {
  @ApiProperty({
    example: 'parent DTO example',
    // type: 'string', // This fixes the issue
  })
  parentKey: ChildDto;
}

@Controller()
export class HelloController {
  constructor(private readonly appService: AppService) {}

  @Get('/hello/world')
  getHello(
    @Param() params: ParentDto,
    @Query() query: ParentDto,
    @Headers() headers: ParentDto,
    @Body() body: ParentDto,
  ) {
    return this.appService.getHello();
  }
}

```

![Image](https://github.com/user-attachments/assets/2754f3ae-294e-4551-9137-f73c29d4168b)

### Minimum reproduction code

<see current behavior>

### Steps to reproduce

<see current behavior>

### Expected behavior

All examples (`@Body`, `@Header`, `@Param`, or `@Query`) use the `ParentDto` example.

### Package version

11.0.3

### NestJS version

11.0.10

### Node.js version

22.5.1

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

kamilmysliwiec:
Let's track this here https://github.com/nestjs/swagger/pull/3782

PR #3883, opened 2026-04-28
fix(response-object-factory): preserve example/examples for built-in scalar response types

yogeshwaran-c:

Bug fix.

When a response decorator is configured with a built-in scalar (or array of scalar) type and an example payload, the example is silently dropped from the generated OpenAPI document.

```ts
@ApiOkResponse({ type: Number, example: 42 })
findAge() { ... }
```

The generated 200 response has no example:

```json
{
  "description": "",
  "content": {
    "application/json": { "schema": { "type": "number" } }
  }
}
```

`ResponseObjectFactory#create` strips `example` and `examples` from the response via `omit(response, exampleKeys)` for the built-in-type branch, but it never re-attaches them to the generated media-type entry. The non-built-in (DTO) branch — handled by `ResponseObjectMapper` — already does the right thing via `pick(response, exampleKeys)` when it constructs the media type. The built-in-type branch is the outlier.

## What is the new behavior?

`example` / `examples` are now forwarded to `MimetypeContentWrapper.wrap` for both the scalar and array-of-scalar built-in-type code paths, mirroring the DTO behaviour. The example ends up next to `schema` inside each `content[<mimetype>]` entry, as required by the OpenAPI spec, and is no longer left behind at the response top level.

```json
{
  "description": "",
  "content": {
    "application/json": {
      "schema": { "type": "number" },
      "example": 42
    }
  }
}
```

## Additional context (optional)

Added a focused unit test for `ResponseObjectFactory` covering scalar `example`, scalar `examples`, the `isArray: true` codepath, and a guard that `example` does not leak back to the top-level response object. The full `swagger-explorer` and `response-object-mapper` suites continue to pass (63 tests).

### DIFF
--- a/e2e/src/cats/cats.controller.ts
+++ b/e2e/src/cats/cats.controller.ts
@@ -243,4 +243,34 @@ export class CatsController {
     }
   })
   rawSchemaResponse() {}
+
+  @Get('scalar-with-example')
+  @ApiResponse({ status: 200, type: Number, example: 42 })
+  scalarWithExample(): number {
+    return 42;
+  }
+
+  @Get('scalar-with-examples')
+  @ApiResponse({
+    status: 200,
+    type: Number,
+    examples: {
+      adult: { value: 5, summary: 'Adult cat age' },
+      kitten: { value: 1, summary: 'Kitten age' }
+    }
+  })
+  scalarWithExamples(): number {
+    return 5;
+  }
+
+  @Get('array-of-scalar-with-example')
+  @ApiResponse({
+    status: 200,
+    type: String,
+    isArray: true,
+    example: ['Mau', 'Persian']
+  })
+  arrayOfScalarWithExample(): string[] {
+    return ['Mau', 'Persian'];
+  }
 }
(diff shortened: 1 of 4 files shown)


kamilmysliwiec:
Could you update e2e tests as well?

yogeshwaran-c:
@kamilmysliwiec done — pushed 1f305dfd which adds e2e coverage for this fix.

Three new endpoints in \`e2e/src/cats/cats.controller.ts\`:
- \`GET /cats/scalar-with-example\` — \`@ApiResponse({ status: 200, type: Number, example: 42 })\`
- \`GET /cats/scalar-with-examples\` — \`@ApiResponse({ ..., examples: { adult, kitten } })\`
- \`GET /cats/array-of-scalar-with-example\` — \`@ApiResponse({ ..., type: String, isArray: true, example: ['Mau', 'Persian'] })\`

A new test in \`e2e/validate-schema.e2e-spec.ts\` (\`should preserve example/examples for built-in scalar response types\`) asserts:
- \`example\` lands at \`responses['200'].content['application/json'].example\` (not at the response top level)
- \`examples\` keyed-object survives intact
- The array path produces \`schema: { type: 'array', items: { type: 'string' } }\` with the example alongside

\`npm run test:e2e -- validate-schema -t "preserve example"\` → 2/2 passing on the new test plus the existing #3335 sibling.

Issue #1665, opened 2021-11-11
Support union type in swagger plugin


RageYL:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Currently, the swagger plugin is unable to to annotate union types properties with @ApiProperty({ enum: ... }), it only works with enum.

Prisma ORM generates union type instead of enum which forces us to annotate all Dto.

### Describe the solution you'd like

Detect union types and generate the corresponding @ApiProperty({ enum: ... }).

Unfortunately, after trying for a few hours, I lack the knowledge and time to submit a working PR.

### Teachability, documentation, adoption, migration strategy

N/A

### What is the motivation / use case for changing the behavior?

Reduce boilerplate and maintainability.

nartc:
Thanks for submitting the issue. Can you provide a sample repo so that I know what I'm working against?

RageYL:
I was testing by declaring the following type, it is used exactly like an enum:
```
export type Status = 'enabled' | 'disabled'
```
and declared an attribute in my Dto with the type above:
```
status: Status
```
Do you still need a repo?

nartc:
Thank you for providing that example. I just take a quick look at the AST with a `TypeAlias` and a property referencing that type:

```ts
export type Status = 'enabled' | 'disabled';

export class Foo {
   status: Status;
}
```

Unfortunately, at the moment the AST doesn't give enough information about the `TypeReference` that `status` is referencing. All it knows is that the type has a name of `Status` without giving any information about `enable` or `disable` (which I had hoped it did :( ). 

That said, I think this cannot be done at the moment.

RageYL:
When I investigated, I was able to get the values by accessing the `type.types` array.

```
console.log(type.types.map(t => typeChecker.typeToString(t)))
```
This prints `['enabled', 'disabled']` for 'status: Status' and `['undefined', 'enabled', 'disabled']` for 'status?: Status'.

nartc:
@RageYL if that's the case, then it can be supported. I haven't looked into the code yet, only tried things out in AST Explorer.

nartc:
@micalevisk this is all I've got from ASTExplorer
![image](https://user-images.githubusercontent.com/25516557/141491754-3f16d863-1305-4d74-9ceb-525b43e93cc5.png)
![image](https://user-images.githubusercontent.com/25516557/141491758-47913323-a7fd-4217-a26f-005939a3553c.png)

[...]

PR #3885, opened 2026-04-30
feat(plugin): auto-generate enum metadata for string and number literal union types

y-hsgw:
## Summary

Revisits #1665.

In issue #1665, the feature was closed with _"There are no plans to support this feature atm."_ However, after investigating the implementation, it appears feasible with a relatively small change. I'd like to propose this as a candidate for support — happy to iterate if the approach needs adjustment.

### What this does

The CLI plugin now automatically infers `enum: [...]` metadata for properties whose type is a string or number literal union — without requiring a manual `@ApiProperty({ enum: [...] })`.

**Before**

```ts
type Color = 'red' | 'green' | 'blue';

export class CatDto {
  // Previously required manual annotation:
  @ApiProperty({ enum: ['red', 'green', 'blue'] })
  color: Color;
}
```

**After**

```ts
type Color = 'red' | 'green' | 'blue';

export class CatDto {
  color: Color;
}
```

Generated metadata: `{ color: { required: true, enum: ["red", "green", "blue"] } }` ✅

### Cases covered

| Type | Generated metadata |
|------|--------------------|
| `'a' \| 'b'` | `enum: ["a", "b"]` |
| `'a' \| 'b' \| null` | `enum: ["a", "b"], nullable: true` |
| `Color` (type alias) | `enum: ["red", "green", "blue"]` |
| `Color \| null` | `enum: ["red", "green", "blue"], nullable: true` |
| `Color?` (optional) | `required: false, enum: [...]` |
| `1 \| 2 \| 3` | `enum: [1, 2, 3]` |
| TypeScript `enum` | unchanged (existing behavior preserved) |

### Implementation

- **`lib/plugin/utils/plugin-utils.ts`**: Added `getStringLiteralUnionValues()` — detects whether a `ts.Type` is a pure string or number literal union (TypeScript enum members are explicitly excluded via `SymbolFlags.EnumMember`), and returns the values along with a `isNullable` flag.
- **`lib/plugin/visitors/model-class.visitor.ts`**:
  - `createTypePropertyAssignments`: treats string/number literal unions like enums (skips emitting `type: () => Object`; emits `nullable: true` when needed)
  - `createEnumPropertyAssignment`: when a literal union is detected, emits `enum: ["a", "b", ...]` or `enum: [1, 2, ...]` as an `ArrayLiteralExpression` instead of a type reference identifier

### Non-goals / out of scope

- Mixed unions like `'a' | 1 | true` — To keep the diff small, and since this is still a proposal, this currently only supports simple unions where all members are string literals or all members are number literals.


kamilmysliwiec:
lgtm, great PR!

Issue #967, opened 2020-09-18
Restrict depth of example values


nick-lehmann:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

The example value for an endpoint is nested as deep as the structure of your entities.

## Expected behavior

After a certain depth, the object shown as an example values should be truncated and only the type shown (if possible).

## Minimal reproduction of the problem with instructions

Should not be needed as the problem is probably dependent on how strong your machine is. However, I can think of an example if desired.

## What is the motivation / use case for changing the behavior?

I am in the process of rewriting an old but quite comphrensive software with around 100 database tables. Thanks to the swagger cli plugin, all Swagger definitions for these tabes are generated automatically. However, there is a group of tables with are highly linked to each others. This causes the example values of some endpoints to become really big (~25k lines per example), making the docs quite slow even on my moderately equipped 2017 MacBook.

## Environment

<pre><code>Nest version: 7.0.8

Others:
@nestjs/swagger: 4.5.12
</code></pre>

kamilmysliwiec:
Would you like to create a PR for this?

nick-lehmann:
Yeah, sure. I have to admit that the code is quite daunting because it is quite much and I didn't figure out at which point the example values are constructed 🤦🏼‍♂️ Can you give me a hint? Anyway, I will give it another try next week..

kamilmysliwiec:
There's currently no bandwidth to fix this issue/implement this feature, unfortunately. Since we don't plan to address this in the foreseeable future, I'm gonna close this issue.

PR #3892, opened 2026-05-08
feat: add exampleMaxDepth to truncate large example values

### LINKED ISSUE #3879: Restrict depth of example values

Reopening https://github.com/nestjs/swagger/issues/967

Maybe somebody has time to implement this?

The example value for an endpoint is nested as deep as the structure of your entities.

After a certain depth, the object shown as an example values should be truncated and only the type shown (if possible).

_No response_

In my current application, generated example value  can reach 45k lines. This temporary freezes openapi documentation page when showing example value

kyungseopk1m:
Closes #3879 (revives #967).

Deeply nested DTOs blow up the generated `example` blob — the issue
reports a 45k-line example that locks up Swagger UI. This adds an
opt-in option to cap how deep example values are emitted.

### Usage

Document-wide:

```ts
SwaggerModule.createDocument(app, config, { exampleMaxDepth: 3 });
```

Per-property override:

```ts
class CreatePostDto {
  @ApiProperty({
    exampleMaxDepth: 1,
    example: { /* very large object */ }
  })
  payload: object;
}
```

Default is `undefined` (no truncation), so existing users see no
change. `0` collapses any non-primitive example to `{}` / `[]`.

### Notes

- Only example values are rewritten. `$ref`, `properties`, `items` and
  the existing circular-dependency guard in `SchemaObjectFactory` are
  untouched.
- The per-property `exampleMaxDepth` is stripped from the emitted
  schema so it never appears in the document.
- Non-plain runtime values (`Date`, `Map`, `Set`, class instances) are
  treated as leaves and passed through as-is, matching the convention
  used by `stripDynamicDefaults`.
- Cycle handling uses a path-local `WeakSet`, so non-cyclic DAGs (the
  same object reachable from multiple siblings) are still trimmed
  fully.
- Scope of this PR is `components.schemas`. Inline examples on
  `RequestBody` / `Response` / `Parameter`, and runtime validation of
  the option value (NaN / negative / Infinity), I'd rather keep as
  separate follow-ups — happy to fold either in here if you prefer.

### Tests

New spec at `test/utils/apply-example-max-depth.util.spec.ts`. Covers
the global limit, per-property override and non-cascade, output-leak
strip, items / allOf / oneOf / anyOf / additionalProperties traversal,
non-cyclic DAGs, `Date` / `Map` / `Set` / class instance pass-through,
cycle guard, and an end-to-end check via `SchemaObjectFactory`. Full
suite 349/349.

### DIFF
--- a/lib/decorators/api-property.decorator.ts
+++ b/lib/decorators/api-property.decorator.ts
@@ -20,6 +20,15 @@ export type ApiPropertyCommonOptions = SchemaObjectMetadata & {
    * @see [Swagger link objects](https://swagger.io/docs/specification/links/)
    */
   link?: () => Type<unknown> | Function;
+  /**
+   * Limits the depth of `example` / `examples` values emitted for this single
+   * schema entry. Overrides `SwaggerDocumentOptions.exampleMaxDepth` for this
+   * property only; sibling and nested schemas are unaffected.
+   *
+   * The metadata is stripped before the document is returned so it never
+   * appears in the generated OpenAPI output.
+   */
+  exampleMaxDepth?: number;
 };
 
 export type ApiPropertyOptions =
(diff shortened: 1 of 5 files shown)


kamilmysliwiec:
lgtm

PR #3897, opened 2026-05-10
fix(response-object-factory): wrap built-in nullable responses in oneOf

yogeshwaran-c:
## Summary

When `@ApiResponse({ type: String, nullable: true })` (or any other built-in scalar) was used, two things went wrong inside `ResponseObjectFactory.create`:

1. The schema emitted under `content[<mime>].schema` ignored the `nullable` flag entirely, so the produced spec did not mark the type as nullable.
2. The `nullable: true` property was leaked onto the OpenAPI ResponseObject root, where it is not a valid field.

`ResponseObjectMapper.toRefObject` / `toArrayRefObject` already handle this case correctly for non-built-in types by wrapping the schema in `oneOf: [<schema>, { type: 'null' }]` and stripping `nullable` from the response root. This PR mirrors that behaviour for the built-in scalar branch (both array and non-array shapes), so the output is consistent across all response shapes and valid OpenAPI.

### Before

```js
factory.create(
  { type: String, description: 'OK', nullable: true },
  ['application/json'], {}, factories
);
```

```jsonc
{
  "description": "OK",
  "nullable": true,                 // <-- leaked to ResponseObject
  "content": {
    "application/json": {
      "schema": { "type": "string" } // <-- nullable not represented
    }
  }
}
```

### After

```jsonc
{
  "description": "OK",
  "content": {
    "application/json": {
      "schema": {
        "oneOf": [{ "type": "string" }, { "type": "null" }]
      }
    }
  }
}
```

The same wrapping is applied for `isArray: true`, producing `oneOf: [{ type: 'array', items: { type: '<scalar>' } }, { type: 'null' }]`. `example` / `examples` siblings are still preserved alongside the wrapped schema.

## Test plan

- [x] Added three unit tests under `test/services/response-object-factory.spec.ts`:
  - scalar built-in + `nullable: true` produces `oneOf` and strips `nullable` from the response root.
  - array built-in + `nullable: true` wraps the array schema in `oneOf` with `{ type: 'null' }`.
  - `example` is still preserved alongside `nullable`.
- [x] `npm test -- test/services/` (64 tests, all pass).
- [x] `npm run test:e2e` (116 tests, all pass).
- [x] `npm run lint` (no new warnings/errors).
- [x] `npm run build` (clean).

### DIFF
--- a/lib/services/response-object-factory.ts
+++ b/lib/services/response-object-factory.ts
@@ -58,30 +58,21 @@ export class ResponseObjectFactory {
         this.swaggerTypesMapper.mapTypeToOpenAPIType(typeName);
 
       const exampleKeys = ['example', 'examples'];
-      if (isArray) {
-        const content = this.mimetypeContentWrapper.wrap(produces, {
-          schema: {
-            type: 'array',
-            items: {
-              type: swaggerType
-            }
-          },
-          ...pick(response, exampleKeys)
-        });
-        return {
-          ...omit(response, exampleKeys),
-          ...content
-        };
-      }
-
+      const baseSchema: SchemaObject = isArray
+        ? { type: 'array', items: { type: swaggerType } }
+        : { type: swaggerType };
+      // Mirror the nullable handling in ResponseObjectMapper so the
+      // "nullable" flag does not leak into the response object root and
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #3899, opened 2026-05-10
fix(plugin): only unwrap Promise/Observable when they are the top-level type

yogeshwaran-c:

Bug fix.

`isPromiseOrObservable` in `lib/plugin/utils/plugin-utils.ts` decides whether the plugin should "unwrap" a return type by doing a substring match for `Promise<` or `Observable<`. Any user-defined class whose name happens to *end* with one of those words therefore matches as well, so the inner type argument is silently substituted for the actual type in the generated metadata.

Example controller:

```ts
class Cat {}

class MyPromise<T> {
  value: T;
}

@Controller('cats')
export class AppController {
  @Get()
  findOne(): MyPromise<Cat> { return null; }
}
```

What the plugin emits today:

```js
openapi.ApiResponse({ status: 200, type: Cat })
```

`Cat` is a completely different class. The same bug affects DTO properties typed as `MyPromise<Cat>` (model-class visitor goes through the same `getTypeReferenceAsString` helper).

Also reproducible with anything ending in `Observable`, e.g. `AsyncObservable<Cat>`, `MyObservable<Cat>`.

## What is the new behavior?

Anchor the regex to the start of the type text (or to a fully qualified `import("...").` prefix) so only true `Promise<...>` / `Observable<...>` references are unwrapped. User-defined classes whose names happen to share the suffix are left alone, and no bogus `type: <inner>` is emitted.

## Additional context

- Added fixture `test/plugin/fixtures/app.controller-promise-prefix.ts` covering all three return types: standard `Promise<Cat>` (still unwraps), `MyPromise<Cat>` (no longer unwraps), `AsyncObservable<Cat>` (no longer unwraps).
- Added matching spec case in `test/plugin/controller-class-visitor.spec.ts`. The new case fails before the fix and passes after.
- All existing plugin tests still pass: `npx vitest run` 331/331; e2e 116/116; `npm run lint` reports no new errors.

### DIFF
--- a/lib/plugin/utils/plugin-utils.ts
+++ b/lib/plugin/utils/plugin-utils.ts
@@ -128,8 +128,18 @@ export function getTypeReferenceAsString(
   }
 }
 
+/**
+ * Returns `true` when the supplied type text refers to a top-level
+ * `Promise<...>` or `Observable<...>` instantiation.
+ *
+ * The match is anchored to the start of the string (or to a fully qualified
+ * `import("...").` prefix) so user-defined classes whose names happen to end
+ * with `Promise` / `Observable` (for example `MyPromise<T>` or
+ * `AsyncObservable<T>`) are not mistakenly unwrapped to their inner type
+ * argument.
+ */
 export function isPromiseOrObservable(type: string) {
-  return type.includes('Promise<') || type.includes('Observable<');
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
lgtm

PR #3911, opened 2026-05-14
fix(swagger-module): return reply from async route handlers

### LINKED ISSUE #3908: Fastify: /docs-json returns 200 with empty body since 11.3.0 (async handler regression)

[Diff 11.2.7...11.3.0 in `lib/swagger-module.ts`](https://github.com/nestjs/swagger/compare/11.2.7...11.3.0#diff-2c4cbf2c7a1eaffe6f4d7e60aebdd8f1) — the JSON, YAML and `swagger-ui-init.js` route handlers were converted from sync to `async` to support an async `patchDocumentOnRequest` hook.

### Versions

`11.2.7` → `11.3.0` (regression introduced); still present in `11.4.2` (latest).

### Describe the regression

After upgrading `@nestjs/swagger` from 11.2.7 to 11.3.0+, Swagger UI renders blank in the browser when using `FastifyAdapter` **together with `@fastify/compress`** (a very common setup).

Symptom (visible only in the browser / with an `Accept-Encoding: br|gzip` request — plain `curl` without compression looks fine):
- `GET /docs-json` → HTTP 200, headers include `content-encoding: br` (or `gzip`), `content-length: 0` → empty body
- `GET /docs/swagger-ui-init.js` → same, empty body
- Static UI assets (`swagger-ui-bundle.js`, `swagger-ui.css`, …) serve correctly because they go through `@fastify/static`, not through the affected route handlers.

The Swagger UI HTML page is served correctly and there are no errors in the browser console — only the JSON document and the init script come back empty, so swagger-ui has nothing to initialize.

**Root cause**

In `lib/swagger-module.ts` (11.3.0+), the JSON, YAML and `swagger-ui-init.js` route handlers are now `async` but still call `res.send(...)` without returning anything:

```ts
httpAdapter.get(normalizeRelPath(options.jsonDocumentUrl), async (req, res) => {
  res.type('application/json');
  const document = getBuiltDocument();
  // ...
  res.send(JSON.stringify(documentToSerialize)); // <-- no return
});
```

When `@fastify/compress` is registered, its `onSend` hook expects to receive the payload through the normal Fastify lifecycle. Async handlers should return the payload (or `reply`) so the hook can pipe it through brotli/gzip. Because the handler resolves to `undefined`, the compression hook serializes `undefined` and emits an empty compressed body, even though `reply.send()` was called manually inside.

Express (`@nestjs/platform-express`) is not affected because its `res.send()` ends the response synchronously without involving any post-handler hook pipeline. The bug is also invisible on Fastify alone (no `@fastify/compress` registered), which is why the unit tests didn't catch it.

The same pattern affects 5 handlers in `lib/swagger-module.ts` (11.4.2 line numbers): 128, 142, 188, 203 — and the existing inner `return res.send(...)` already present on lines 123 and 137 (inside the `patchDocumentOnRequest` branches) confirms the intended pattern.

### Minimum reproduction code

```ts
import fastifyCompress from '@fastify/compress';
import { NestFactory } from '@nestjs/core';
import { FastifyAdapter, NestFastifyApplication } from '@nestjs/platform-fastify';
import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
import { AppModule } from './app.module';

async function bootstrap() {
  const app = await NestFactory.create<NestFastifyApplication>(AppModule, new FastifyAdapter());
  await app.register(fastifyCompress); // <-- the trigger
  const config = new DocumentBuilder().setTitle('App').setVersion('1.0').build();
  const document = SwaggerModule.createDocument(app, config);
  SwaggerModule.setup('docs', app, document);
  await app.listen(3000);
}
bootstrap();
```

```
$ curl -is -H 'Accept-Encoding: br' http://localhost:3000/docs-json
HTTP/1.1 200 OK
content-type: application/json; charset=utf-8
vary: accept-encoding
content-encoding: br
content-length: 0

# (empty body)
```

Confirmed fix (patched `node_modules/@nestjs/swagger/dist/swagger-module.js` locally, added `return` before each `res.send(...)` in the 4 affected async handlers, restarted the app):

```
$ curl -s -H 'Accept-Encoding: br' http://localhost:3000/docs-json | wc -c
11854   # brotli-compressed payload, decompresses to the full 198 KB OpenAPI doc
```

Removing `@fastify/compress` (or downgrading to `@nestjs/swagger@11.2.7`) also restores correct behavior.

### Expected behavior

`GET /docs-json`, `/docs-yaml`, and `/docs/swagger-ui-init.js` should return their bodies regardless of `Accept-Encoding`, matching the 11.2.7 behavior and the Express adapter's behavior on 11.3.0+.

### Suggested fix

Return the result of `res.send(...)` from each async handler in `lib/swagger-module.ts`:

```ts
async (req, res) => {
  res.type('application/json');
  // ...
  return res.send(JSON.stringify(documentToSerialize));
}
```

Affected lines in 11.4.2: **128**, **142**, **188**, **203**. The `patchDocumentOnRequest` branches (lines 123, 137) already use this pattern.

### Other

- `@nestjs/swagger`: 11.3.0 → 11.4.2 (broken with `@fastify/compress`), 11.2.7 (working)
- `@nestjs/platform-fastify`: 11.1.20
- `fastify`: 5.8.5
- `@fastify/compress`: 8.3.1
- Node: 24.x
- OS: macOS 24.6.0
- Package manager: pnpm 10.33.0

tibohaffner:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When `@fastify/compress` is registered on a `NestFastifyApplication`, Swagger UI renders blank because `GET /docs-json`, `GET /docs-yaml`, and `GET /docs/swagger-ui-init.js` all return HTTP 200 with an **empty body** as soon as the client sends an `Accept-Encoding: br|gzip` header (which every modern browser does by default).

In 11.3.0, the JSON, YAML and `swagger-ui-init.js` route handlers in `lib/swagger-module.ts` were converted from sync to `async` to support an async `patchDocumentOnRequest` hook, but the bare `res.send(...)` calls were left without a `return`. The async function therefore resolves to `undefined`. When `@fastify/compress` attaches its `onSend` hook to compress the response, it serializes `undefined` and emits an empty compressed body — even though `reply.send()` was called manually inside the handler.

The HTML page and static assets (served through `@fastify/static`) are unaffected, which is why the page loads but the UI never initializes.

Plain `curl` requests without `Accept-Encoding` look fine, which made the regression invisible to the existing test suite.

Closes #3908

## What is the new behavior?

The 4 affected async handlers now `return res.send(...)`, mirroring the pattern already used in the `patchDocumentOnRequest` branches. The async function resolves to the reply, the `onSend` hook receives the actual payload, and compression works correctly.

Affected handlers (line numbers from this PR):
- `swagger-module.ts:212` — `swagger-ui-init.js` (primary route)
- `swagger-module.ts:244` — `swagger-ui-init.js` (trailing-slash route)
- `swagger-module.ts:326` — `/docs-json`
- `swagger-module.ts:353` — `/docs-yaml`

The sync HTML handler (`swaggerUiHtml`, line 276) was left unchanged because it is not affected.

### Tests

Added a new `describe('with @fastify/compress registered', ...)` block in `e2e/fastify.e2e-spec.ts` covering `Accept-Encoding: br` and `gzip` against `/apidoc-json`, `/apidoc-yaml`, and `/apidoc/swagger-ui-init.js`. Verified that all 4 new tests fail on master (each asserts body length > 0 — currently 0) and pass with this fix.

Added `@fastify/compress` as a `devDependency` (already a peer of `@nestjs/platform-fastify` consumers).

Full test suites green: 122 e2e passed, 368 unit passed, `tsc -p tsconfig.build.json` clean.

- [ ] Yes
- [x] No

The change preserves the existing behavior for callers not using `@fastify/compress` (since `reply.send()` already terminates the response) and is identical to what the `patchDocumentOnRequest` branches do today.

## Other information

The fix was validated end-to-end in a real project (NestJS 11.1.20 + `@nestjs/platform-fastify` 11.1.20 + `@fastify/compress` 8.3.1 + `fastify` 5.8.5) by patching `node_modules` locally and confirming that:
- `curl -s -H 'Accept-Encoding: br' /api/docs-json | wc -c` returns ~12 KB of brotli payload (was 0 before)
- the decompressed body matches the full 198 KB OpenAPI document
- Swagger UI renders normally in Safari, Chrome and Firefox

### DIFF
--- a/e2e/fastify.e2e-spec.ts
+++ b/e2e/fastify.e2e-spec.ts
@@ -686,4 +686,66 @@ describe('Fastify Swagger', () => {
       await app.close();
     });
   });
+
+  describe('with @fastify/compress registered', () => {
+    const SWAGGER_RELATIVE_URL = '/apidoc';
+
+    beforeEach(async () => {
+      const fastifyCompress = (await import('@fastify/compress')).default;
+      await app.register(fastifyCompress as any);
+
+      const swaggerDocument = SwaggerModule.createDocument(
+        app,
+        builder.build()
+      );
+      SwaggerModule.setup(SWAGGER_RELATIVE_URL, app, swaggerDocument);
+
+      await app.init();
+      await app.getHttpAdapter().getInstance().ready();
+    });
+
+    afterEach(async () => {
+      await app.close();
+    });
+
+    it('should serve the JSON definition with brotli compression', async () => {
+      const response = await request(app.getHttpServer())
+        .get(`${SWAGGER_RELATIVE_URL}-json`)
+        .set('Accept-Encoding', 'br');
+
+      expect(response.status).toEqual(200);
+      expect(response.headers['content-encoding']).toEqual('br');
+      expect(Object.keys(response.body).length).toBeGreaterThan(0);
+    });
+
+    it('should serve the JSON definition with gzip compression', async () => {
+      const response = await request(app.getHttpServer())
+        .get(`${SWAGGER_RELATIVE_URL}-json`)
+        .set('Accept-Encoding', 'gzip');
+
+      expect(response.status).toEqual(200);
+      expect(response.headers['content-encoding']).toEqual('gzip');
+      expect(Object.keys(response.body).length).toBeGreaterThan(0);
+    });
+
+    it('should serve the YAML definition with brotli compression', async () => {
+      const response = await request(app.getHttpServer())
+        .get(`${SWAGGER_RELATIVE_URL}-yaml`)
+        .set('Accept-Encoding', 'br');
+
+      expect(response.status).toEqual(200);
+      expect(response.headers['content-encoding']).toEqual('br');
+      expect(response.text.length).toBeGreaterThan(0);
+    });
+
+    it('should serve swagger-ui-init.js with brotli compression', async () => {
+      const response = await request(app.getHttpServer())
+        .get(`${SWAGGER_RELATIVE_URL}/swagger-ui-init.js`)
+        .set('Accept-Encoding', 'br');
+
+      expect(response.status).toEqual(200);
+      expect(response.headers['content-encoding']).toEqual('br');
+      expect(response.text.length).toBeGreaterThan(0);
+    });
+  });
 });
--- a/lib/swagger-module.ts
+++ b/lib/swagger-module.ts
@@ -209,7 +209,7 @@ export class SwaggerModule {
           swaggerUiInitJS = buildSwaggerInitJS(document, swaggerOptions);
         }
 
-        res.send(swaggerUiInitJS);
+        return res.send(swaggerUiInitJS);
       }
     );
 
@@ -241,7 +241,7 @@ export class SwaggerModule {
             swaggerUiInitJS = buildSwaggerInitJS(document, swaggerOptions);
           }
 
-          res.send(swaggerUiInitJS);
+          return res.send(swaggerUiInitJS);
         }
       );
     } catch {
@@ -323,7 +323,7 @@ export class SwaggerModule {
               )
             : document;
 
-          res.send(JSON.stringify(documentToSerialize));
+          return res.send(JSON.stringify(documentToSerialize));
         }
       );
     }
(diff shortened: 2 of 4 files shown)


kamilmysliwiec:
lgtm

PR #3914, opened 2026-05-15
feat(swagger): support raw component schemas

### LINKED ISSUE #3886: Allow registering raw top-level component schemas

## Feature request

Allow registering **raw top-level component schemas** (esp. `oneOf` + `discriminator` with `mapping`) without having to express them as a class. Equivalently: a class-level decorator, e.g. `@ApiDiscriminator({ propertyName, mapping })`, that emits the class as a `oneOf+discriminator` entry under `components.schemas`.

## Why

`@nestjs/swagger` reflects classes into `{ type: "object", properties }` schemas. Combinators (`oneOf`/`anyOf`/`allOf`) are honored only when applied to **fields** via `@ApiProperty({ oneOf, discriminator })` — that emits the union **inline** at the operation's request/response, **not** as a top-level `components.schemas.*` entry.

Confirmed in `lib/services/schema-object-factory.ts` (combinators only checked while iterating model properties):

```ts
const schemaCombinators = ['oneOf', 'anyOf', 'allOf'];
// ... only honored on properties of a class, not the class itself
```

Workarounds today, all unsatisfying:

1. `@ApiBody({ schema: { oneOf, discriminator } })` + `@ApiExtraModels(...children)` → inline only, duplicated per operation, can't be `$ref`d, and downstream tooling that walks **only** `components.schemas` for discriminator preprocessing (graphql-mesh, some SDK generators) silently ignores it.
2. Make a marker class with no fields and `@ApiExtraModels` it → emits `{ type: "object" }`, useless for unions.
3. Mutate the document after `SwaggerModule.createDocument()` (what we ended up doing) → out-of-band, fragile.

## Concrete impact

Any service that returns a polymorphic array where all branches share the same envelope. Example: an endpoint returns `events: EventDto[]` where each event is one of `EventCreatedDto | EventUpdatedDto | EventDeletedDto`, all sharing common fields (`id`, `kind`, `timestamp`, `payload`). Without a top-level `oneOf+discriminator+mapping` component schema, downstream consumers (graphql-mesh, code generators that walk `components.schemas` only) cannot resolve the concrete branch — they fall back to structural matching, which always picks the first branch. Result: every `EventUpdated` is mis-typed as `EventCreated` end-to-end.

## Repro

Minimal repo: any class-based DTO returning a polymorphic array. Today the only path to a top-level discriminator entry is post-hoc document mutation.

## Related

- #3549 — different bug (oneOf downgraded to `$ref: Object`), already fixed in #3781.
- #3160 — `@ApiSchema` inheritance, naming-only.
- #2180 — swagger-ui rendering of inline discriminators (different layer).

None of these address top-level `components.schemas` registration of unions.

## Environment

- `@nestjs/swagger`: 7.4.2 (also missing in `8.x`/`11.x` — checked source).
- Node 22.

Happy to send a PR if there's appetite for the `@ApiDiscriminator` shape.

## Proposed API

A class-level decorator:

```ts
@ApiDiscriminator({
  propertyName: 'kind',
  mapping: {
    created: EventCreatedDto,
    updated: EventUpdatedDto,
    deleted: EventDeletedDto,
  },
})
export class EventDto {}
```

Emits to `components.schemas.EventDto`:

```yaml
EventDto:
  oneOf:
    - $ref: '#/components/schemas/EventCreatedDto'
    - $ref: '#/components/schemas/EventUpdatedDto'
    - $ref: '#/components/schemas/EventDeletedDto'
  discriminator:
    propertyName: kind
    mapping:
      created: '#/components/schemas/EventCreatedDto'
      updated: '#/components/schemas/EventUpdatedDto'
      deleted: '#/components/schemas/EventDeletedDto'
```

Used as the type for properties:

```ts
@ApiProperty({ type: () => EventDto, isArray: true })
events: EventDto[];
```

Resolves to `events: { type: array, items: { $ref: '#/components/schemas/EventDto' } }` — a single canonical, reusable, dedup'd top-level schema.

Alternative API: extend `SwaggerDocumentOptions.extraModels` (or add a new option) to accept raw OpenAPI schemas keyed by name:

```ts
SwaggerModule.createDocument(app, config, {
  extraSchemas: {
    EventDto: { oneOf: [...], discriminator: { ... } },
  },
});
```

_No response_

Any service that returns a polymorphic array where all branches share the same envelope. Example: an endpoint returns `events: EventDto[]` where each event is one of `EventCreatedDto | EventUpdatedDto | EventDeletedDto`, all sharing common fields (`id`, `kind`, `timestamp`, `payload`). Without a top-level `oneOf+discriminator+mapping` component schema, downstream consumers (graphql-mesh, code generators that walk `components.schemas` only) cannot resolve the concrete branch — they fall back to structural matching, which always picks the first branch. Result: every `EventUpdated` is mis-typed as `EventCreated` end-to-end.

yudin-s:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

`@ApiSchema()` metadata is always merged into a generated object schema. Even when a class-level schema declares `oneOf`, `anyOf`, or `allOf`, the resulting `components.schemas` entry still gets generated object scaffolding such as `type: 'object'` and `properties: {}`.

This prevents using a class-level placeholder as a raw top-level component schema for cases like `oneOf` plus `discriminator.mapping`.

Issue Number: Fixes #3886

## What is the new behavior?

`@ApiSchema()` now accepts the full `SchemaObject` shape. When a class-level schema declares `oneOf`, `anyOf`, or `allOf`, `SchemaObjectFactory` registers it as a raw top-level component schema instead of wrapping it in generated object metadata.

This lets users define component schemas such as:

```ts
@ApiSchema({
  name: 'Pet',
  oneOf: [
    { $ref: '#/components/schemas/Cat' },
    { $ref: '#/components/schemas/Dog' },
  ],
  discriminator: {
    propertyName: 'type',
    mapping: {
      cat: '#/components/schemas/Cat',
      dog: '#/components/schemas/Dog',
    },
  },
})
class PetDto {}
```

The existing object-schema behavior is preserved for regular DTOs and for `@ApiSchema()` options without schema combinators.

- [ ] Yes
- [x] No

## Other information

Local checks:

- `npm test -- test/services/schema-object-factory.spec.ts`
- `npm test -- test/explorer/swagger-explorer.spec.ts`
- `npm run build`
- `npm run lint` (passes with existing warnings outside this change)
- `npm exec prettier -- --check lib/decorators/api-schema.decorator.ts lib/services/schema-object-factory.ts test/services/schema-object-factory.spec.ts test/explorer/swagger-explorer.spec.ts`
- `git diff --check`

Note: local Node is `v20.18.3`, while some dependencies now warn for `^20.19.0 || >=22.12.0`. I installed missing optional native packages for `rolldown` and `oxlint` without changing the lockfile to run the checks.

### DIFF
--- a/e2e/validate-schema.e2e-spec.ts
+++ b/e2e/validate-schema.e2e-spec.ts
@@ -1,10 +1,12 @@
-import { INestApplication } from '@nestjs/common';
+import { Controller, Get, INestApplication } from '@nestjs/common';
 import { NestFactory } from '@nestjs/core';
 import { writeFileSync } from 'fs';
 import { OpenAPIV3 } from 'openapi-types';
 import { join } from 'path';
 import type { BaseIssue, BaseSchema } from 'valibot';
 import {
+  ApiResponse,
+  ApiSchema,
   DocumentBuilder,
   getSchemaPath,
   OpenAPIObject,
@@ -27,6 +29,25 @@ import { toJsonSchema } from '@valibot/to-json-schema';
 import type { ZodType } from 'zod';
 import SwaggerParser = require('@apidevtools/swagger-parser');
 
+@ApiSchema({
+  name: 'RawUnionResponse',
+  oneOf: [{ type: 'string' }, { type: 'number' }],
+  description: 'Raw union schema component'
+})
+class RawUnionResponse {}
+
+@Controller('raw-component-schemas')
+class RawComponentSchemaController {
+  @Get('union')
+  @ApiResponse({
+    status: 200,
+    type: RawUnionResponse
+  })
+  getRawUnionResponse() {
+    return null;
+  }
+}
+
 function asResponseObject(
   response: ResponseObject | { $ref: string } | undefined
 ) {
@@ -46,7 +67,7 @@ describe('Validate OpenAPI schema', () => {
       {
         module: class {},
         imports: [ApplicationModule],
-        controllers: [ExpressController]
+        controllers: [ExpressController, RawComponentSchemaController]
       },
       {
         logger: false
@@ -370,6 +391,23 @@ describe('Validate OpenAPI schema', () => {
   it('should add extension to root', () => {
     const document = SwaggerModule.createDocument(app, options, documentOptions);
     expect(document['x-test']).toEqual({ test: 'test' });
+  });
+
+  it('should register raw component schema with combinators', () => {
+    const document = SwaggerModule.createDocument(app, options, documentOptions);
+
+    const response = asResponseObject(
+      document.paths['/api/raw-component-schemas/union']['get']['responses'][
+        '200'
+      ]
+    );
+    expect(response.content!['application/json'].schema).toEqual({
+      $ref: '#/components/schemas/RawUnionResponse'
+    });
+    expect(document.components?.schemas?.RawUnionResponse).toEqual({
+      oneOf: [{ type: 'string' }, { type: 'number' }],
+      description: 'Raw union schema component'
+    });
   });
 
   it('should add extension to info', () => {
--- a/lib/decorators/api-schema.decorator.ts
+++ b/lib/decorators/api-schema.decorator.ts
@@ -1,17 +1,12 @@
 import { DECORATORS } from '../constants.js';
-import { SchemaObjectMetadata } from '../interfaces/schema-object-metadata.interface.js';
+import { SchemaObject } from '../interfaces/open-api-spec.interface.js';
 import { createClassDecorator } from './helpers.js';
 
-export interface ApiSchemaOptions extends Pick<SchemaObjectMetadata, 'name'> {
+export interface ApiSchemaOptions extends SchemaObject {
   /**
    * Name of the schema.
    */
   name?: string;
-
-  /**
-   * Description of the schema.
-   */
-  description?: string;
 }
 
 /**
--- a/lib/services/schema-object-factory.ts
+++ b/lib/services/schema-object-factory.ts
@@ -38,6 +38,8 @@ import { ParamWithTypeMetadata } from './parameter-metadata-accessor.js';
 import { StandardSchemaOpenApiConverter } from './standard-schema-openapi.converter.js';
 import { SwaggerTypesMapper } from './swagger-types-mapper.js';
 
+const schemaCombinators = ['oneOf', 'anyOf', 'allOf'] as const;
+
 export class SchemaObjectFactory {
   private readonly standardSchemaOpenApiConverter =
     new StandardSchemaOpenApiConverter(this.standardSchemaConverter);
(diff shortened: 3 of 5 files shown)


kamilmysliwiec:
could you please target v12.0.0?

yudin-s:
@kamilmysliwiec  we don't have v12.0.0 on this repo 

Rebased and retargeted this PR to `chore/esm-migration`, which appears to be the active v12 branch at the moment (I couldn't find a separate `v12.0.0` branch in the repository).

kamilmysliwiec:
could you update e2e tests as well?

yudin-s:
Done!

dogemad:
I think this is solving a named schema registration problem rather than a type modeling problem.

Using an empty class as a placeholder is misleading because it makes `PetDto` look like a real DTO or a common base type, while it is actually only being used as a runtime token for registering an OpenAPI component schema.

Because `PetDto` is an empty class, it does not establish a meaningful type relationship with `CatDto` or `DogDto`, and it does not model `CatDto | DogDto` in TypeScript.

For example:

```ts
@ApiSchema({
  name: 'Pet',
  oneOf: [
    { $ref: getSchemaPath(CatDto) },
    { $ref: getSchemaPath(DogDto) },
  ],
  discriminator: {
    propertyName: 'type',
    mapping: {
      cat: getSchemaPath(CatDto),
      dog: getSchemaPath(DogDto),
    },
  },
})
class PetDto {}
```

The actual application type still needs to be declared separately:

```ts
type PetDto = CatDto | DogDto;
```

Would it make more sense to expose a helper for registering a named `SchemaObject` directly?

```ts
const PetApiSchema = defineApiSchema('Pet', {
  oneOf: [
    { $ref: getSchemaPath(CatDto) },
    { $ref: getSchemaPath(DogDto) },
  ],
  discriminator: {
    propertyName: 'type',
    mapping: {
      cat: getSchemaPath(CatDto),
      dog: getSchemaPath(DogDto),
    },
  },
});
```

The helper could accept the standard OpenAPI `SchemaObject` shape rather than introducing a union-specific abstraction. That would preserve the existing OpenAPI model while also supporting primitives, arrays, enums, `allOf`, `anyOf`, `oneOf`, `not`, and other valid schema constructs.

There is already a similar precedent in `@nestjs/graphql`, where `createUnionType()` is used to register union schema metadata instead of requiring an empty placeholder class that appears to represent an application type.

A similar helper here would keep the responsibilities clear:

* TypeScript types describe the application model.
* `SchemaObject` describes the OpenAPI schema.
* The helper registers the schema under `components.schemas`.
* No placeholder class is required purely for schema registration.

This seems more consistent with the existing NestJS approach to schema-level constructs and avoids overloading DTO classes with schema-registration semantics.

kamilmysliwiec:
1. allOf silently discards the class own properties
2. in check treats an explicitly-undefined combinator as raw (example `@ApiSchema({ name: 'Undef', oneOf: undefined, description: 'd' })`)
3. ApiSchemaOptions now extends the whole SchemaObject, which invites raw declarations that aren't combinators (and raw mode only keys on the three combinators)

PR #3725, opened 2026-02-05
feat(swagger): add OpenAPI 3.2 hierarchical tags support

apt-bh:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

This PR implements the feature I requested in #3607. I wanted to give it a try myself.

Currently, `TagObject` only supports `name`, `description`, and `externalDocs`. OpenAPI 3.2 introduced hierarchical tags via `parent` and `kind` fields (doc [here](https://spec.openapis.org/oas/v3.2.0.html#fixed-fields-18) ), but there's no way to use them with `@nestjs/swagger`.

## What is the new behavior?

- `TagObject` interface now includes optional `parent` and `kind` fields
- `DocumentBuilder.addTag()` accepts an optional 4th parameter for hierarchy options
- `@ApiTags` decorator accepts object form `{ name, parent?, kind? }` alongside strings
- New `ApiTagOptions` type is exported for TypeScript users

```typescript
// Define tag hierarchy at document level
const config = new DocumentBuilder()
  .addTag('Animals', 'All animal operations')
  .addTag('Cats', 'Cat operations', undefined, { parent: 'Animals' })
  .addTag('Dogs', 'Dog operations', undefined, { parent: 'Animals', kind: 'navigation' })
  .build();

// Use @ApiTags with object form for hierarchy
@ApiTags({ name: 'Cats', parent: 'Animals' })
@Controller('cats')
export class CatsController {}

// Or mix strings and objects
@ApiTags('Dogs', { name: 'Puppies', parent: 'Dogs' })
@Controller('puppies')
export class PuppiesController {}

// Existing string syntax remains fully supported
@ApiTags('Cats')
@Controller('cats')
export class CatsController {}
```

[...]

PR #3924, opened 2026-05-20
feat(swagger): warn when @ApiTags receives hierarchy fields

### LINKED ISSUE #3922: @ApiTags({ name, parent, kind }) silently drops parent and kind
## Is there an existing issue for this?

## Current behavior

PR #3725 made `@ApiTags()` accept an `ApiTagOptions` object form so it can co-exist with strings:

```ts
@ApiTags({ name: 'Cats', parent: 'Animals', kind: 'navigation' })
class CatsController {}
```

However, `parent` and `kind` are silently discarded by the decorator implementation — only `name` is preserved. The PR's own test suite codifies this behavior at [`test/decorators/api-use-tags.decorator.spec.ts`](https://github.com/nestjs/swagger/blob/master/test/decorators/api-use-tags.decorator.spec.ts):

```ts
@ApiTags({ name: 'Cats', parent: 'Animals' })
objectWithParent() {}

// asserts:
expect(tags).toEqual(['Cats']);
```

The architecture supports this drop today: `SwaggerScanner.scanApplication()` does not aggregate tag definitions into `document.tags` — the root-level `tags` array is populated only by `DocumentBuilder.addTag()`. The decorator-level metadata has nowhere to land.

Net effect for consumers: passing `@ApiTags({ name: 'Pets', parent: 'Animals' })` is exactly equivalent to `@ApiTags('Pets')`, but the type signature suggests otherwise.

## Expected behavior

One of:

1. **Propagate** — `@ApiTags(ApiTagOptions)` aggregates definitions into `document.tags`, deduping by `name` with entries declared via `DocumentBuilder.addTag()`. This requires:
   - Decorator persists full options (in addition to/instead of just names).
   - A new global tag-definitions explorer.
   - Scanner aggregates and returns them in `document.tags`.
   - `SwaggerModule.createDocument` merges scanner-collected tags with config tags.

2. **Reject / warn** — emit a runtime warning (or compile-time error via type narrowing) when hierarchy fields are set on `@ApiTags`, redirecting users to `DocumentBuilder.addTag()`. Less surprising than silent drop.

Option (1) matches what the type signature already promises and is what most users would expect. Option (2) keeps the current architecture intact.

Happy to open a PR once a direction is decided. Related: #3921 (adds `summary` field at the builder level — same OpenAPI 3.2 work, deliberately scoped narrower because of this open question).

## Package version
@nestjs/swagger >= 11.4.0

## What is the motivation / use case for changing the behavior?

OpenAPI 3.2 hierarchical tags are useful for organizing large APIs. Today users have to declare every tag's `parent`/`kind` via `DocumentBuilder.addTag()` in `main.ts`, even though `@ApiTags()` *looks* like it accepts the same options.

frbuceta:
## PR Type

- [x] Feature
- [ ] Bug fix
- [ ] Other

`@ApiTags` accepts the object form `{ name, parent, kind }` (added in #3725, released in v11.4.0), but it normalizes every entry to its `name`, **silently discarding `parent` and `kind`**:

```ts
const normalizedTags = tags.map((tag) =>
  typeof tag === 'string' ? tag : tag.name
);
```

The decorator-level hierarchy metadata never reaches `document.tags`, so the OpenAPI 3.2 tag hierarchy is lost with no feedback to the user. Root-level tag hierarchy is only honored when declared via `DocumentBuilder.addTag(name, description, { parent, kind })`.

Issue: #3922

## What is the new behavior?

When `parent` or `kind` is set on an `@ApiTags` object argument, the decorator now emits a runtime warning that points users to `DocumentBuilder.addTag()` — the supported way to define a root-level tag hierarchy.

```
[ApiTags] Tag "Sessions" was declared with hierarchy fields (parent/kind) on
@ApiTags, which are dropped. Define them via DocumentBuilder.addTag("Sessions",
description, { parent, kind }) so they reach the root-level tags of the document.
```

This implements the resolution agreed in #3922 ("Reject / warn"). String arguments and object arguments that only carry `name` are unaffected — no warning is emitted.

- [ ] Yes
- [x] No

Tag normalization is unchanged; only a diagnostic warning is added.

## Other information

Closes #3922. Tests added for the warn / no-warn cases in `test/decorators/api-use-tags.decorator.spec.ts`.

### DIFF
--- a/lib/decorators/api-use-tags.decorator.ts
+++ b/lib/decorators/api-use-tags.decorator.ts
@@ -14,11 +14,7 @@ export function ApiTags(...tags: (string | ApiTagOptions)[]) {
       return tag;
     }
     if (tag.parent !== undefined || tag.kind !== undefined) {
-      logger.warn(
-        `Tag "${tag.name}" was declared with hierarchy fields (parent/kind) on @ApiTags, ` +
-          `which are dropped. Define them via DocumentBuilder.addTag("${tag.name}", ` +
-          `description, { parent, kind }) so they reach the root-level tags of the document.`
-      );
+      logger.warn(`Tag "${tag.name}" includes hierarchy fields (parent/kind) defined via @ApiTags, but those values are ignored there. Define them with DocumentBuilder.addTag("${tag.name}", description, { parent, kind }) to include them in the document's root-level tags.`);
     }
     return tag.name;
   });
--- a/lib/decorators/api-use-tags.decorator.ts
+++ b/lib/decorators/api-use-tags.decorator.ts
@@ -1,13 +1,26 @@
+import { Logger } from '@nestjs/common';
 import { DECORATORS } from '../constants';
 import { ApiTagOptions } from '../interfaces/open-api-spec.interface';
 import { createMixedDecorator } from './helpers';
+
+const logger = new Logger('ApiTags');
 
 /**
  * @publicApi
  */
 export function ApiTags(...tags: (string | ApiTagOptions)[]) {
-  const normalizedTags = tags.map((tag) =>
-    typeof tag === 'string' ? tag : tag.name
-  );
+  const normalizedTags = tags.map((tag) => {
+    if (typeof tag === 'string') {
+      return tag;
+    }
+    if (tag.parent !== undefined || tag.kind !== undefined) {
+      logger.warn(
+        `Tag "${tag.name}" was declared with hierarchy fields (parent/kind) on @ApiTags, ` +
+          `which are dropped. Define them via DocumentBuilder.addTag("${tag.name}", ` +
+          `description, { parent, kind }) so they reach the root-level tags of the document.`
+      );
+    }
+    return tag.name;
+  });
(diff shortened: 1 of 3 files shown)


lib/decorators/api-use-tags.decorator.ts
```diff
@@ -1,13 +1,26 @@
...
+  const normalizedTags = tags.map((tag) => {
+    if (typeof tag === 'string') {
+      return tag;
+    }
+    if (tag.parent !== undefined || tag.kind !== undefined) {
+      logger.warn(
+        `Tag "${tag.name}" was declared with hierarchy fields (parent/kind) on @ApiTags, ` +
+          `which are dropped. Define them via DocumentBuilder.addTag("${tag.name}", ` +
+          `description, { parent, kind }) so they reach the root-level tags of the document.`
+      );
```
kamilmysliwiec:
```suggestion
      logger.warn(`Tag "${tag.name}" includes hierarchy fields (parent/kind) defined via @ApiTags, but those values are ignored there. Define them with DocumentBuilder.addTag("${tag.name}", description, { parent, kind }) to include them in the document's root-level tags.`);
```

PR #3938, opened 2026-05-29
fix(plugin): preserve package import paths

### LINKED ISSUE #3937: CLI plugin emits filesystem-relative require() to workspace package internals instead of the source import path

### Current behavior

When I reference a sibling package `@repro/shared` in monorepo:

```ts
// note: our case is importing from nested folder, but I have tested this with importing from package root and the results is same
import { ItemStatus } from "@repro/shared/messages";

export class ItemDto {
  id!: string;
  status!: ItemStatus;
}
```
It gets compiled into relative path:

```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.ItemDto = void 0;
const openapi = require("@nestjs/swagger");
class ItemDto {
    id;
    status;
        return {
            id:     { required: true, type: () => String },
            status: { required: true, enum: require("../../../../packages/shared/dist/messages/item").ItemStatus }
        };
    }
}
exports.ItemDto = ItemDto;
```

which then breaks at runtime, because `@repro/shared` package is after build in `node_modules/@repro/shared` and not at `../../../../packages/shared`

### Minimum reproduction code

https://github.com/datageo-cz/datageo-swagger-plugin-error-reproduction

### Steps to reproduce

> Hope you dont mind the reproduction repo was AI generated (the issue and the original code was not), but I have gone through it manually and checked the issue there is present and correct.

1. build project
    ```bash
    pnpm install
    pnpm turbo build --filter @repro/api
    ```
2. Inspect the compiled output and note the `require("../../../../packages/shared/dist/messages/item")` inside the `_OPENAPI_METADATA_FACTORY` body.

### Expected behavior

I would expect @nestjs/swagger plugin to keep the nonrelative import path from @repro/shared as it is, thus keeping the ability to run the built project.

### Package version

11.2.6

### NestJS version

11.1.9

### Node.js version

22.16.0

- [ ] macOS
- [ ] Windows
- [x] Linux

### Other

very similar issue was mentioned in https://github.com/nestjs/swagger/issues/3453#issuecomment-3352670679 by @mrVito. The issue was closed as the OP's problem was little bit different and solved.

yudin-s:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When the CLI plugin sees a DTO property typed from a workspace package import such as `@repro/shared/messages`, TypeScript may resolve the type to the package physical output path. The generated `_OPENAPI_METADATA_FACTORY()` can then emit a filesystem-relative `require()` into `packages/shared/dist/...`, which is not stable at runtime after the package is installed from `node_modules`.

Issue Number: Fixes #3937

## What is the new behavior?

The model visitor now preserves explicit package import specifiers for imported property/enum types before normalizing paths. Package-style references keep emitting package-based `require()` calls, while relative and built-in/bare module behavior stays unchanged.

A regression fixture covers a monorepo-style DTO importing an enum from `@repro/shared/messages` and asserts the emitted metadata uses `require("@repro/shared/messages").ItemStatus` instead of a resolved sibling package path.

- [ ] Yes
- [x] No

## Other information

Docs were not updated because this is a CLI plugin code-generation bug fix with regression coverage.

Local validation:

```bash
npm test -- test/plugin/plugin-utils.spec.ts test/plugin/model-class-visitor.spec.ts
npm test -- test/plugin
npm run build -- --pretty false
npm exec -- prettier --check lib/plugin/utils/plugin-utils.ts lib/plugin/visitors/model-class.visitor.ts test/plugin/plugin-utils.spec.ts test/plugin/model-class-visitor.spec.ts test/plugin/fixtures/package-imports/apps/api/src/item.dto.ts test/plugin/fixtures/package-imports/packages/shared/dist/messages/item.d.ts test/plugin/fixtures/package-imports/tsconfig.json
npm run lint
git diff --check
```

Note: local install required restoring optional native packages for Vitest/oxlint on macOS; no package manifest or lockfile changes are included.

### DIFF
--- a/lib/plugin/utils/plugin-utils.ts
+++ b/lib/plugin/utils/plugin-utils.ts
@@ -129,6 +129,122 @@ export function getTypeReferenceAsString(
 }
 
 /**
+ * Best-effort extraction of a package-level import specifier from a type node.
+ * For a type imported as `import { ItemStatus } from '@repro/shared/messages';`,
+ * this returns `import("@repro/shared/messages").ItemStatus`.
+ */
+export function getImportedTypeReference(
+  typeNode: ts.TypeNode | undefined,
+  typeChecker: ts.TypeChecker
+): string | undefined {
+  const typeReferenceNode = getImportableTypeReferenceNode(typeNode);
+  if (!typeReferenceNode) {
+    return;
+  }
+
+  const typeName = typeReferenceNode.typeName;
+  if (ts.isQualifiedName(typeName)) {
+    const namespaceSymbol = typeChecker.getSymbolAtLocation(typeName.left);
+    const namespaceImport = namespaceSymbol?.declarations?.find(
+      ts.isNamespaceImport
+    );
+
+    if (
+      !namespaceImport ||
+      !ts.isImportDeclaration(namespaceImport.parent.parent) ||
+      !ts.isStringLiteral(namespaceImport.parent.parent.moduleSpecifier)
+    ) {
+      return;
+    }
+
+    const moduleSpecifier = namespaceImport.parent.parent.moduleSpecifier.text;
+    if (!isPackageSpecifier(moduleSpecifier)) {
+      return;
+    }
+    return `import("${moduleSpecifier}").${typeName.right.text}`;
+  }
+
+  const symbol = typeChecker.getSymbolAtLocation(typeName);
+  const declaration = symbol?.declarations?.find(
+    (item): item is ts.ImportClause | ts.ImportSpecifier =>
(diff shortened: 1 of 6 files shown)


SmallhillCZ:
whoah! hoped this would be simpler :) thanks for working on this!

SmallhillCZ:
@yudin-s @kamilmysliwiec hi, is there an ETA for merging? thanks

SmallhillCZ:
@kamilmysliwiec hi, is this closed as not planned or as will be fixed elsewhere?

PR #3945, opened 2026-06-01
fix(mimetype-content-wrapper): clone object per mimetype to avoid shared references

yogeshwaran-c:
## Summary

`MimetypeContentWrapper#wrap` builds the OpenAPI `content` object by mapping each media type to `removeUndefinedKeys(obj)` — but it reuses the **same object reference** for every media type and passes the caller's object directly into `removeUndefinedKeys` (which mutates in place).

Two consequences for responses/request bodies declared with multiple media types (e.g. `@ApiProduces('application/json', 'application/xml')` or multiple `@ApiConsumes`):

1. **Aliased entries** — every media-type entry in `content` points to the *same* object instance, so mutating one media type's schema later silently mutates all of them.
2. **Source mutation** — `removeUndefinedKeys` strips keys from the caller's original object rather than a copy.

Reproduction (current behavior):

```ts
const { content } = new MimetypeContentWrapper().wrap(
  ['application/json', 'application/xml'],
  { schema: { type: 'string' } }
);

content['application/json'] === content['application/xml']; // true (bug)
content['application/json'].schema.type = 'mutated';
content['application/xml'].schema.type; // 'mutated' (leaked)
```

The fix clones the object per media type with lodash `cloneDeep` (already used elsewhere in the codebase) so each entry is independent and the caller's source object is left untouched.

## Test plan

Added `test/services/mimetype-content-wrapper.spec.ts` covering:

- wrapping under each provided mimetype
- stripping `undefined` keys
- **distinct object/schema references between media types** (fails before the fix)
- **caller's source object is not mutated** (fails before the fix)

```
npx vitest run test/services            # 71 passed (incl. new spec)
npm run lint                            # exit 0
npm run build                           # exit 0
```

### DIFF
--- a/lib/services/mimetype-content-wrapper.ts
+++ b/lib/services/mimetype-content-wrapper.ts
@@ -1,3 +1,4 @@
+import { cloneDeep } from 'lodash';
 import { ContentObject } from '../interfaces/open-api-spec.interface';
 import { removeUndefinedKeys } from '../utils/remove-undefined-keys';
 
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #3948, opened 2026-06-01
fix(plugin): emit object/null values for @example JSDoc tag

yogeshwaran-c:
## Summary

The CLI plugin silently drops a model property's entire metadata when its `@example` JSDoc tag holds an **object** (or an array of objects).

`createLiteralFromAnyValue` (in `lib/plugin/utils/ast-utils.ts`) only handled arrays and primitives. For a plain object it fell through to `createPrimitiveLiteral`, which returns `undefined`. That `undefined` was then passed to `factory.createPropertyAssignment('example', undefined)`, which throws. The `ModelClassVisitor` wraps property inspection in a `try/catch` that swallows the error and returns the node unchanged, so the **whole property** disappears from the generated `_OPENAPI_METADATA_FACTORY`.

Example that triggered the bug:

```ts
export class Dto {
  /** @example [{ "id": 1 }, { "id": 2 }] */
  items: object[]; // <- this property was entirely omitted from metadata
}
```

### Fix

Extend `createLiteralFromAnyValue` to also handle plain objects (recursively, emitting an object literal) and `null`. Arrays and primitives behave exactly as before.

## Test plan

- Added two properties (`testObjectExample`, `testObjectArrayExample`) with object-valued `@example` tags to the existing `create-cat-alt2.dto` plugin fixture and updated the expected transpiled `_OPENAPI_METADATA_FACTORY` output.
- The new assertions **fail on master** (the object-example properties are missing from the emitted metadata) and **pass with this fix**.
- Full plugin suite green: `vitest run test/plugin` (58 passed). `oxlint lib/` clean (no new warnings) and `tsc -p tsconfig.build.json` builds.

### DIFF
--- a/lib/plugin/utils/ast-utils.ts
+++ b/lib/plugin/utils/ast-utils.ts
@@ -349,9 +349,23 @@ export function createLiteralFromAnyValue(
   factory: ts.NodeFactory,
   item: unknown
 ) {
-  return Array.isArray(item)
-    ? factory.createArrayLiteralExpression(
-        item.map((item) => createLiteralFromAnyValue(factory, item))
+  if (Array.isArray(item)) {
+    return factory.createArrayLiteralExpression(
+      item.map((item) => createLiteralFromAnyValue(factory, item))
+    );
+  }
+  if (item === null) {
+    return factory.createNull();
+  }
+  if (typeof item === 'object') {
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
could you make all PRs you just created target v12.0.0 branch?

PR #3957, opened 2026-06-06
feat(plugin): generate additionalProperties for Record/index-signature types

y-hsgw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

With the CLI plugin enabled, a property typed as `Record<string, V>` (or an inline index signature `{ [key: string]: V }`) does not produce a useful schema. The generated `@ApiProperty` metadata resolves to a bare `{ type: 'object' }` with no `additionalProperties`, so the value type is lost.

For example:

```ts
export class Dto {
  attributes: Record<string, string>;
}
```

produces `{ type: 'object' }`, which `openapi-typescript` turns into `attributes: Record<string, never>` (an empty object), and an inline `{ [key: string]: V }` property is dropped from the metadata entirely.

Issue Number: N/A

## What is the new behavior?

The plugin now detects `Record<string, V>` and inline index signatures and emits flat `type: 'object'` + `additionalProperties` describing the value type:

- `Record<string, string>` → `{ type: 'object', additionalProperties: { type: 'string' } }`
- `Record<string, number>` → `additionalProperties: { type: 'number' }`
- `Record<string, boolean>` → `additionalProperties: { type: 'boolean' }`
- `Record<string, V[]>` → `additionalProperties: { type: 'array', items: ... }`
- nested `Record<string, Record<string, string>>` → recurses
- `Record<string, any>` / `unknown` → `additionalProperties: true`
- class/DTO value types fall back to `additionalProperties: true` (a proper `$ref` requires extending model exploration and is intentionally out of scope here)

`null`/`undefined` unions (`Record<string, string> | null`, optional `?`) compose with the existing nullable handling.

The metadata is emitted flat (not wrapped in a lazy `type: () => ({ ... })` resolver) so the runtime does not treat the object literal as a nested model.

- [ ] Yes
- [x] No

A property previously emitted as `{ type: 'object' }` for `Record<string, V>` now additionally carries `additionalProperties`. Existing properties that explicitly set `@ApiProperty({ type: 'object', additionalProperties: ... })` are unchanged (the plugin skips properties that already define `type`).

## Other information

Added a plugin fixture covering the primitive/array/nested/nullable/optional cases and `SchemaObjectFactory` tests verifying the generated metadata resolves to the expected schema.


kamilmysliwiec:
lgtm

PR #3964, opened 2026-06-15
feat(plugin): infer ApiParam enum from @Param literal-union types

y-hsgw:
## PR Type
- [x] Feature

The Swagger CLI plugin does not infer enum metadata for controller route
parameters. For a `@Param('code')` whose type is a string- or number-literal
union (e.g. `type Color = 'red' | 'blue'`), the generated OpenAPI spec does not
document the allowed values, so developers must repeat them in a hand-written
`@ApiParam({ name, enum: [...] })`.

```ts
type Color = 'red' | 'blue';

@Controller('items')
export class ItemsController {
  // The plugin emits no enum metadata for `code`; the only way to document
  // the allowed values today is to repeat them in an explicit @ApiParam.
  @Get(':code')
  findOne(@Param('code') code: Color) {}
}
```

Issue Number: N/A

## What is the new behavior?

The plugin now auto-generates `@ApiParam({ name, enum: [...] })` for each
`@Param('name')` parameter whose type resolves to a string- or number-literal
union. The enum values are taken from the parameter's **type**, so no extra
decorator is needed.

```ts
type Color = 'red' | 'blue';

@Controller('items')
export class ItemsController {
  // The plugin now generates @ApiParam({ name: 'code', enum: ['red', 'blue'] }).
  @Get(':code')
  findOne(@Param('code') code: Color) {}
}
```

- Supports literal-union type aliases (`type Color = 'red' | 'blue'`), inline
  unions (`'a' | 'b' | 'c'`), and numeric unions (`1 | 2 | 3`).
- An existing `@ApiParam` for the same name takes precedence and is left
  untouched. If any `@ApiParam` on the method cannot be reliably introspected
  (no string-literal `name`), generation is skipped for the whole method to
  avoid duplicate metadata.
- A bare `@Param()` (whole params object) and non-literal-union types
  (e.g. `string`) produce no `@ApiParam`.

This mirrors the existing `@ApiQuery` auto-generation. The shared logic for
reading existing decorator names and named param-decorator arguments is
extracted into `ast-utils` (`collectExistingApiParamNames`,
`getNamedParamDecoratorArg`) and reused by both the `@ApiQuery` and `@ApiParam`
generators.

- [x] No

## Other information

Tests covering literal-union type aliases, inline string/number unions,
existing `@ApiParam` precedence, plain (non-enum) types, and bare `@Param()`
were added to `test/plugin/controller-class-visitor.spec.ts`.


kamilmysliwiec:
lgtm

PR #3798, opened 2026-04-02
fix(plugin): normalize workspace package import paths in metadata generator

maruthang:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

`PluginMetadataGenerator` (via `ModelClassVisitor` and `ControllerClassVisitor`) computes relative paths from the source directory to the physical `node_modules/` location of workspace packages. This produces import paths like `../node_modules/@amk/utils/src/dto/order.dto` in the generated metadata, which causes TypeScript TS2742 errors at compile time.

Issue Number: #3453

## What is the new behavior?

A `normalizePackagePath` helper is extracted in `plugin-utils.ts`. It detects when a resolved path passes through `node_modules/` and strips everything up to that boundary, returning a bare package specifier (e.g., `@amk/utils/src/dto/order.dto`). Both `ModelClassVisitor` and `ControllerClassVisitor` now use this helper in their `collectedMetadata()` methods, so generated metadata always contains correct importable package paths.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

7 new unit tests added in `test/plugin/normalize-package-path.spec.ts` covering bare package paths, scoped packages, deep paths, and paths that should not be modified. All 152 tests pass.

PR #4008, opened 2026-07-15
fix(plugin): infer response type for models imported from workspace p…

### LINKED ISSUE #4006: CLI plugin drops inferred response models imported from a workspace package

### Current behavior

This is adjacent to #3937 / #3938, but is a different code path.

#3938 preserves package imports for DTO property metadata. The controller visitor still silently omits an inferred response model when it comes from a workspace package.

### Current behavior

Given a workspace package:

```ts
// @repro/shared
export class SharedResponseDto {
  // The best ID!
  @IsUUID()
  id!: string;
}

and an API controller:

```ts
import { SharedResponseDto } from '@repro/shared';

@Get()
get(): Promise<SharedResponseDto> {
  return { id: uuid() };
}
```
  
with the CLI plugin enabled, the generated controller contains: `openapi.ApiResponse({ status: 200 })`.

The OpenAPI operation consequently has no response content schema.

`ControllerClassVisitor#createTypePropertyAssignment()` resolves the payload as something like:

```ts
import("/absolute/path/node_modules/@repro/shared/dist/index").SharedResponseDto
```

and then returns early because the type name contains `node_modules`.

I checked and for our case FOR SURE the DTO’s `_OPENAPI_METADATA_FACTORY()` works fine once Swagger receives the class reference; it is only the inferred controller response reference that is lost.

### Minimum reproduction code

https://gist.github.com/TheCoreMan/bdac33634090a67e686c0e0f126ddfc7

### Steps to reproduce

_No response_

### Expected behavior

Preserve the source package specifier and emit a response type, e.g.:

```ts
  openapi.ApiResponse({
    status: 200,
    type: require('@repro/shared').SharedResponseDto,
  })
```

### Suggested direction

We introduced this patch locally with `pnpm patch`:

```diff
diff --git a/dist/plugin/visitors/controller-class.visitor.js b/dist/plugin/visitors/controller-class.visitor.js
index 11987703e5692e26d4f2794fe34891acaefde8b2..c4a358cba699679e0da05e7687170a00d5d90521 100644
--- a/dist/plugin/visitors/controller-class.visitor.js
+++ b/dist/plugin/visitors/controller-class.visitor.js
@@ -365,7 +365,13 @@ class ControllerClassVisitor extends abstract_visitor_1.AbstractFileVisitor {
         if (!typeReferenceDescriptor.typeName) {
             return undefined;
         }
-        if (typeReferenceDescriptor.typeName.includes('node_modules')) {
+        if (typeReferenceDescriptor.typeName.includes('node_modules') &&
+            !typeReferenceDescriptor.typeName.includes('node_modules/@opsin/shared/')) {
             return undefined;
         }
         const identifier = (0, type_reference_to_identifier_util_1.typeReferenceToIdentifier)(typeReferenceDescriptor, hostFilename, options, factory, type, this._typeImports);
```

So maybe being able to configure `node_modules` paths that we'd like in that if statement?

Tested with @nestjs/swagger 11.4.1, Nest 11.1.19, TypeScript 6.0.3, and pnpm workspaces on macOS.

### Package version

11.4.1

### NestJS version

11.1.19

### Node.js version

v24

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

thegoldenflow:
The controller visitor bailed whenever the inferred response type resolved through `node_modules` (e.g. a pnpm-symlinked workspace package), emitting `ApiResponse` without a `type` and leaving the operation with no response schema. Remove that guard so control falls through to the existing `typeReferenceToIdentifier` -> `replaceImportPath` -> `normalizePackagePath` chain, aligning the controller response path with the model/property path (#3798).

Fixes #4006

## PR Checklist

- [x] Tests for the changes have been added (mechanism-level; see *Other information* re: coverage scope)
- [ ] Docs have been added / updated — N/A (internal plugin behavior)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

Issue Number: #4006

When a controller method's return type is a DTO imported from a workspace package that resolves through `node_modules` (pnpm symlink), the CLI plugin resolves the type name as `import("/abs/path/node_modules/@scope/pkg/dist/index").Dto`, hits the `typeName.includes('node_modules')` guard in `ControllerClassVisitor#createTypePropertyAssignment()`, and returns early. The generated code is `openapi.ApiResponse({ status: 200 })` with no `type`, so the OpenAPI operation has no response content schema. The DTO's own `_OPENAPI_METADATA_FACTORY` is unaffected — only the inferred controller response reference is lost.

## What is the new behavior?

The guard is removed, so the resolved type flows into `typeReferenceToIdentifier` -> `replaceImportPath` -> `normalizePackagePath`, which strips the `node_modules` prefix and emits the package specifier. The generated response now carries the type, e.g. `openapi.ApiResponse({ status: 200, type: require('@scope/pkg').Dto })`, and the operation gets its response schema. This matches how the model/property path already handles workspace-package imports (#3798).

- [ ] Yes
- [x] No

Additive to the generated OpenAPI document (response schemas that were previously absent now appear). See *Other information* for the behavioral scope.

## Other information

**Scope / tradeoff.** Removing the guard means the controller response path now emits a `type` for *any* return type that resolves through `node_modules`, not only symlinked workspace packages — e.g. `StreamableFile`, `Buffer`, or rxjs types would surface a response schema where previously none was emitted. This is the same tradeoff already accepted for the model/property path in #3798; the blast radius is somewhat larger here since controller return types more often reference framework/third-party types. If you'd prefer to avoid broadening the default, I'm happy to gate this behind a new `PluginOptions` flag (default off) so workspace-package response types are only referenced when explicitly enabled — just let me know.

**Testing.** Added a #4006-shaped case to `normalize-package-path.spec.ts` covering the specifier normalization. The transpile-based plugin unit harness can't reproduce a real pnpm `node_modules` symlink layout, so end-to-end behavior was verified manually against the reporter's minimal reproduction ([<PASTE GIST URL FROM #4006>](https://gist.github.com/TheCoreMan/bdac33634090a67e686c0e0f126ddfc7)): the generated controller emits the response `type` as expected. I can follow up with a runtime integration test that materializes a physical `node_modules/@scope/pkg` in a temp dir and asserts the emitted type, if committed end-to-end coverage is preferred.

### DIFF
--- a/lib/plugin/visitors/controller-class.visitor.ts
+++ b/lib/plugin/visitors/controller-class.visitor.ts
@@ -701,9 +701,6 @@ export class ControllerClassVisitor extends AbstractFileVisitor {
     if (!typeReferenceDescriptor.typeName) {
       return undefined;
     }
-    if (typeReferenceDescriptor.typeName.includes('node_modules')) {
-      return undefined;
-    }
     const identifier = typeReferenceToIdentifier(
       typeReferenceDescriptor,
       hostFilename,
--- a/test/plugin/normalize-package-path.spec.ts
+++ b/test/plugin/normalize-package-path.spec.ts
@@ -46,4 +46,11 @@ describe('normalizePackagePath', () => {
       normalizePackagePath('../../../packages/product-warehouse/dist/index')
     ).toBe('../../../packages/product-warehouse/dist/index');
   });
+
+  // #4006: mechanism-layer only — does not exercise the controller guard removal.
+  it('should strip node_modules prefix for a workspace-package response type (#4006)', () => {
+    expect(
+      normalizePackagePath('../node_modules/@repro/shared/dist/index')
+    ).toBe('@repro/shared/dist');
+  });
 });


kamilmysliwiec:
it converts a missing response schema into an **app-boot crash** for a common subset of the very case it targets

PR #4014, opened 2026-07-17
fix(plugin): preserve inferred responses for enum error statuses

### LINKED ISSUE #4009: enum-based @ApiResponse status prevents automatic success response generation

- [ ] I have read the whole migration guide

#3867 or #3803

### Versions

11.2.3->11.4.4

### Describe the regression

After upgrading to `@nestjs/swagger` `11.4.4`, automatically generated success responses disappeared from Swagger for controller methods that only define explicit error responses.

The issue appears to be caused by the new success-response detection logic in the Swagger plugin.

When an `@ApiResponse` uses an enum reference such as `HttpStatus.BAD_REQUEST`, the plugin does not resolve the enum value. Instead, it appears to treat the response as an explicit success response and skips generating the inferred `200` or `201` response.

### Minimum reproduction code

```ts
import { Controller, Get, HttpStatus } from '@nestjs/common';
import { ApiResponse } from '@nestjs/swagger';

@Controller('example')
export class ExampleController {
  @Get()
  @ApiResponse({
    status: HttpStatus.BAD_REQUEST,
    description: 'Bad request',
  })
  getExample(): ExampleDto {
    return {
      value: 'test',
    };
  }
}
```

### Expected behavior

Swagger should generate:

* the inferred `200` success response with the `ExampleDto` schema
* the explicitly defined `400` error response

### Actual behavior

Only the explicitly defined `400` response is generated.

The inferred `200` response is missing.

### Workaround

Replacing the enum reference with a numeric literal appears to avoid the issue:

```ts
@ApiResponse({
  status: 400,
  description: 'Bad request',
})
```

### Suspected cause

The regression seems related to the logic introduced in `11.4.2`, specifically `isSuccessOrRedirectApiResponseArg`.

The check appears to correctly handle numeric and string literals, but not enum references such as:

```ts
HttpStatus.BAD_REQUEST
```

When the value cannot be resolved, it is treated as a successful response, which causes the plugin to skip automatic success-response generation.

GiHoon1123:
Fixes #4009

### Problem

When `@ApiResponse` uses an enum status such as `HttpStatus.BAD_REQUEST`, the Swagger plugin does not resolve the enum value and treats it as a success response.

As a result, the automatically generated `200` response is omitted.

### Fix

Pass the TypeScript `TypeChecker` to the status check and resolve enum members with `getConstantValue()`.

This preserves the existing behavior for numeric statuses and ensures that:

- enum error statuses keep the inferred success response
- enum success statuses suppress the inferred success response

### Tests

Added regression coverage for:

- `HttpStatus.BAD_REQUEST` with an inferred `200` response
- `HttpStatus.CREATED` suppressing the inferred `200` response

Verified with:

- `pnpm test`
- `pnpm run build`
- `pnpm run lint`

### DIFF
--- a/lib/plugin/visitors/controller-class.visitor.ts
+++ b/lib/plugin/visitors/controller-class.visitor.ts
@@ -188,7 +188,7 @@ export class ControllerClassVisitor extends AbstractFileVisitor {
         const decoratorName = getDecoratorName(item);
         // Error factories (4xx/5xx) must not suppress the auto-inferred 2xx.
         if (decoratorName === ApiResponse.name) {
-          return this.isSuccessOrRedirectApiResponseArg(item);
+          return this.isSuccessOrRedirectApiResponseArg(item, typeChecker);
         }
         const statusNameMatch = decoratorName.match(/^Api(.+)Response$/);
         if (!statusNameMatch) return false;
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #4025, opened 2026-07-24
feat(decorators): add reusable union schema factory

dogemad:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] CI related changes

Reusable schemas composed with OpenAPI `oneOf` or `anyOf` must be declared manually. There is no public factory that creates a named, type-safe union schema which can be reused as a model or referenced through helpers such as `refs()`.

Issue Number: #3886

## What is the new behavior?

This PR adds `createUnionApiSchema()`, a public factory for declaring reusable OpenAPI `oneOf` and `anyOf` component schemas from model classes and supported built-in types.

```ts
class CatDto {
  @ApiProperty()
  name: string;
}

class DogDto {
  @ApiProperty()
  breed: string;
}

const Pet = createUnionApiSchema({
  name: 'Pet',
  oneOf: [CatDto, DogDto]
});

type Pet = InferUnionApiSchema<typeof Pet>;
```

The generated component is equivalent to:

```yaml
Pet:
  oneOf:
    - $ref: '#/components/schemas/CatDto'
    - $ref: '#/components/schemas/DogDto'
```

The public API includes:

- `createUnionApiSchema`
- `InferUnionApiSchema`
- `UnionApiSchema`
- `UnionApiSchemaOptions`

The implementation also:

- uses the raw combinator support provided through `ApiSchema` / `API_SCHEMA`, without introducing a separate union-specific metadata key or a second union branch in `SchemaObjectFactory`;
- rejects empty unions at runtime as well as at compile time;
- rejects unsupported members, including `Array`, instead of emitting permissive schemas;
- reuses `SwaggerTypesMapper` for built-in type mappings;
- deduplicates repeated union members and detects conflicting component names;
- validates generated component names, including unsafe names such as `__proto__`;
- assigns the configured component name to the generated metadata host so `refs(Pet)` resolves to `#/components/schemas/Pet`;
- keeps the factory under `lib/decorators/`, since its result is a metadata host rather than an extendable DTO class.

## Review feedback addressed

- Reverted unrelated Prettier reformatting in `e2e/src/cats/cats.controller.ts`.
- Added runtime validation for empty unions.
- Added union-member deduplication.
- Reused `SwaggerTypesMapper` and removed permissive `Array` handling.
- Moved the factory out of `lib/type-helpers/`.
- Removed the unrelated `generateSchema()` custom-name fix from this PR so it can be handled independently.
- Built the union factory on the shared `API_SCHEMA` raw-schema foundation rather than adding `API_UNION_SCHEMA`.

The documentation PR is still required before release. The landing order with #3914 should also be coordinated so both PRs use one shared raw-schema foundation.

- [ ] Yes
- [x] No

## Other information

Validation completed locally:

- Build passes.
- Unit tests pass: 46 test files, 507 tests.
- E2E tests pass: 5 test files, 127 tests.
- Lint passes with no errors (existing warnings only).
- Prettier and staged/unstaged diff checks pass.
- The generated OpenAPI document passes Swagger Parser validation.


kamilmysliwiec:
Opus review below:

Thanks for this - the implementation is clean and the type-level work (`const Members`, `MemberResult` inference, the "member can't share the union's schema name" guard) is well done. I verified locally that custom `@ApiSchema({ name })` on members resolves correctly, nested unions work, discriminator mappings auto-register models outside `oneOf`, and `cloneDeep` correctly protects the stored metadata.

Before we can merge, a few things:

### 1. Overlap with #3914

#3914 also fixes #3886, and both PRs add an early-return branch to `exploreModelSchema` that registers a raw combinator schema. If both land we end up with two near-identical mechanisms (`API_SCHEMA` with combinators vs. `API_UNION_SCHEMA`). Let's settle on one foundation first — `createUnionApiSchema` could be built on top of the raw-schema support rather than introducing a second metadata key.

### 2. Unrelated reformatting

About 90 lines in `e2e/src/cats/cats.controller.ts` are pure prettier rewrapping of the Zod `standardBodyRichSchema` / `standardResponseRichSchema` definitions, unrelated to this feature. Please revert those — they inflate the diff and will conflict.

### 3. Empty unions are only rejected at compile time

```ts
createUnionApiSchema({ name: 'E', oneOf: [] as any })
// → { "E": { "oneOf": [] } }   // invalid OpenAPI, no error
```

`NonEmptyUnionMembers` catches this in TS, but the factory already does a runtime check for `Function` members, so this should be guarded at runtime too.

### 4. Smaller items

- **Duplicate members aren't deduped** — `oneOf: [X, X]` emits the same `$ref` twice. `models` is a `Set`, but the `oneOf` array isn't.
- **`getBuiltInSchema` duplicates `SwaggerTypesMapper`** — two sources of truth that will drift. Also `Array → { type: 'array', items: {} }` accepts anything; consider dropping `Array` rather than emitting a permissive schema.
- **`lib/type-helpers/` is the wrong home** — everything there returns a real class you can extend and instantiate; this returns a metadata host that must not be. `lib/decorators/` or its own module fits better.
- **Split out the `generateSchema` fix** — using the name returned by `exploreModelSchema` instead of `target.name` is a real fix for custom `@ApiSchema({ name })` lookups and unrelated to unions. Its own PR would let it land independently and get its own changelog entry.
- **Docs** — this adds public API (`createUnionApiSchema`, `InferUnionApiSchema`, `UnionApiSchema`, `UnionApiSchemaOptions`) with a const-plus-type-alias pattern people will get wrong. Needs a docs.nestjs.com PR before release.

dogemad:
@kamilmysliwiec 

Thanks for the detailed review — I’ve addressed the code feedback:
- Removed the separate API_UNION_SCHEMA mechanism. createUnionApiSchema() now builds on the shared API_SCHEMA raw-combinator path.
- Reverted the unrelated Prettier reformatting.
- Added runtime validation for empty unions and invalid members.
- Deduplicated oneOf members and guarded component-path collisions.
- Reused SwaggerTypesMapper and rejected Array.
- Moved the factory to lib/decorators/.
- Removed the unrelated generateSchema() fix.
- Added coverage for nested unions, discriminator targets, raw combinator handling, and invalid OpenAPI edge cases.
All unit tests, e2e tests, build, lint, and OpenAPI validation pass. The remaining non-code item is the docs.nestjs.com PR, which I’ll link separately. For #3914, the implementation now uses the same API_SCHEMA foundation, but the landing order still needs to be coordinated.

Issue #4063, opened 2026-08-19
Schema properties keep the 3.0 nullable keyword in documents declared as OpenAPI 3.1


simonboeuf-kpler:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

`DocumentBuilder.setOpenAPIVersion('3.1.0')` sets the document version to 3.1.0.
In 3.1, a nullable property is a type union: `{ "type": ["string", "null"] }`.
The schema object factory cannot emit this union for a property:

```ts
// deletedAt: string | null
// @ApiProperty({ type: String })
//   emits: { "type": "string" }                    (nullability absent)
// @ApiProperty({ type: String, nullable: true })
//   emits: { "type": "string", "nullable": true }  (removed 3.0 keyword)
// expected in a 3.1 document:
//          { "type": ["string", "null"] }
```

OpenAPI 3.1 uses JSON Schema 2020-12, which removed the `nullable` keyword. JSON Schema ignores unknown keywords. A strict 3.1 consumer therefore reads both outputs as a plain non-nullable string. The nullability is lost, and no tool gives a warning. Code generators that follow 3.1 strictly produce clients without the null case. These clients fail on the first real null value.

The CLI plugin gives the same output. For a property typed `string | null`, the plugin adds `nullable: true` (`model-class.visitor.ts`, `isNullableUnion`).

An explicit union in the options is a possible workaround: `@ApiProperty({ anyOf: [{ type: 'string' }, { type: 'null' }] })`. But the type is restated by hand for each field, and for a property typed `string | null` the CLI plugin injects a sibling `nullable: true` next to the authored union.

The library already emits the correct 3.1 spelling for nullable inline responses: since #3897, the response object factory produces `oneOf: [<schema>, { "type": "null" }]` and drops the keyword. Schema properties need the same treatment. Today one document can mix both dialects, and the declared version changes neither output.

### Minimum reproduction code

https://gist.github.com/simonboeuf-kpler/4cd47410fe874b57376be6c0f888b4cf

### Steps to reproduce

1. Download the four gist files into an empty folder.
2. Install and run:

```bash
npm install
npm run repro
```

3. Open `openapi.json`. The document declares `"openapi": "3.1.0"`, and `UserDto`
   shows both variants:
   - `deletedAt` (nullability from the TS type only) emits `{ "type": "string" }` — the null case is absent.
[...]

PR #4064, opened 2026-08-19
fix(): rewrite nullable schemas for OpenAPI 3.1 documents

official-burak:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

\`DocumentBuilder.setOpenAPIVersion('3.1.0')\` sets \`openapi: 3.1.0\`, but schema properties still emit the OpenAPI 3.0 \`nullable\` keyword:

\`\`\`ts
@ApiProperty({ type: String, nullable: true })
// { \"type\": \"string\", \"nullable\": true }
\`\`\`

OpenAPI 3.1 uses JSON Schema 2020-12, which removed \`nullable\`. Strict 3.1 consumers therefore treat the property as a non-nullable string.

Issue Number: #4063

## What is the new behavior?

When the document version is 3.1 or later, \`SwaggerModule.createDocument\` rewrites:

- \`{ type: \"string\", nullable: true }\` to \`{ type: [\"string\", \"null\"] }\`
- nested property schemas the same way
- \`{ nullable: true, allOf: [{ \$ref }] }\` to \`{ oneOf: [allOf, { type: \"null\" }] }\`

3.0 documents are unchanged.

- [ ] Yes
- [x] No

Only 3.1+ documents change, and only to match the 3.1 schema language.

## Other information

Unit tests cover scalar unions, type arrays, nested properties, and nullable \`$ref\`/\`allOf\` schemas.

### DIFF
--- a/lib/utils/rewrite-nullable-for-oas31.util.ts
+++ b/lib/utils/rewrite-nullable-for-oas31.util.ts
@@ -1,0 +1,42 @@
+/**
+ * OpenAPI 3.1 uses JSON Schema 2020-12, which dropped the `nullable` keyword.
+ * Convert 3.0-style `{ type: T, nullable: true }` into 3.1 unions.
+ */
+export function rewriteNullableForOas31(value: unknown): void {
+  if (Array.isArray(value)) {
+    for (const item of value) {
+      rewriteNullableForOas31(item);
+    }
+    return;
+  }
+  if (!value || typeof value !== 'object') {
+    return;
+  }
+
+  const obj = value as Record<string, unknown>;
+  for (const nested of Object.values(obj)) {
+    rewriteNullableForOas31(nested);
+  }
+
+  if (obj.nullable !== true) {
+    return;
+  }
+  delete obj.nullable;
+
+  if (typeof obj.type === 'string') {
+    obj.type = [obj.type, 'null'];
+    return;
+  }
+  if (Array.isArray(obj.type)) {
+    if (!obj.type.includes('null')) {
+      obj.type = [...obj.type, 'null'];
+    }
+    return;
+  }
+
+  const rest = { ...obj };
+  for (const key of Object.keys(obj)) {
+    delete obj[key];
+  }
+  obj.oneOf = [rest, { type: 'null' }];
+}
--- a/test/utils/rewrite-nullable-for-oas31.util.spec.ts
+++ b/test/utils/rewrite-nullable-for-oas31.util.spec.ts
@@ -1,0 +1,64 @@
+import { rewriteNullableForOas31 } from '../../lib/utils/rewrite-nullable-for-oas31.util';
+
+describe('rewriteNullableForOas31', () => {
+  it('converts a scalar type plus nullable into a type union', () => {
+    const schema: any = { type: 'string', nullable: true };
+
+    rewriteNullableForOas31(schema);
+
+    expect(schema).toEqual({ type: ['string', 'null'] });
+    expect(schema.nullable).toBeUndefined();
+  });
+
+  it('appends null to an existing type array', () => {
+    const schema: any = { type: ['string', 'number'], nullable: true };
+
+    rewriteNullableForOas31(schema);
+
+    expect(schema.type).toEqual(['string', 'number', 'null']);
+  });
+
+  it('rewrites nested property schemas', () => {
+    const schemas: any = {
+      User: {
+        type: 'object',
+        properties: {
+          deletedAt: { type: 'string', format: 'date-time', nullable: true },
+          name: { type: 'string' }
+        }
+      }
+    };
+
+    rewriteNullableForOas31(schemas);
+
+    expect(schemas.User.properties.deletedAt).toEqual({
+      type: ['string', 'null'],
+      format: 'date-time'
+    });
+    expect(schemas.User.properties.name).toEqual({ type: 'string' });
+  });
+
+  it('represents a nullable $ref/allOf schema as oneOf with null', () => {
+    const schema: any = {
+      nullable: true,
+      allOf: [{ $ref: '#/components/schemas/Profile' }]
+    };
+
+    rewriteNullableForOas31(schema);
+
+    expect(schema).toEqual({
+      oneOf: [
+        { allOf: [{ $ref: '#/components/schemas/Profile' }] },
+        { type: 'null' }
+      ]
+    });
+  });
+
+  it('leaves 3.0-compatible schemas without nullable unchanged', () => {
+    const schema: any = { type: 'string' };
+
+    rewriteNullableForOas31(schema);
+
+    expect(schema).toEqual({ type: 'string' });
+  });
+});
--- a/lib/swagger-module.ts
+++ b/lib/swagger-module.ts
@@ -19,6 +19,7 @@ import {
 import { assignTwoLevelsDeep } from './utils/assign-two-levels-deep.js';
 import { getGlobalPrefix } from './utils/get-global-prefix.js';
 import { isOas31OrLater } from './utils/is-oas31-or-later.util.js';
+import { rewriteNullableForOas31 } from './utils/rewrite-nullable-for-oas31.util.js';
 import { normalizeRelPath } from './utils/normalize-rel-path.js';
 import { resolvePath } from './utils/resolve-path.util.js';
 import { validateGlobalPrefix } from './utils/validate-global-prefix.util.js';
@@ -76,19 +77,23 @@ export class SwaggerModule {
       ...documentWithoutWebhooks
     };
 
-    if (shouldIncludeWebhooks) {
-      return {
-        ...baseDocument,
-        ...(mergedWebhooks ? { webhooks: mergedWebhooks } : {})
-      };
+    const documentForVersion = shouldIncludeWebhooks
+      ? {
+          ...baseDocument,
+          ...(mergedWebhooks ? { webhooks: mergedWebhooks } : {})
+        }
+      : {
+          ...baseDocument,
+          paths: webhookPaths
+            ? assignTwoLevelsDeep({}, baseDocument.paths || {}, webhookPaths)
+            : baseDocument.paths
+        };
+
+    if (isOas31OrLater(documentForVersion.openapi ?? '3.0.0')) {
+      rewriteNullableForOas31(documentForVersion);
     }
 
-    return {
-      ...baseDocument,
-      paths: webhookPaths
-        ? assignTwoLevelsDeep({}, baseDocument.paths || {}, webhookPaths)
-        : baseDocument.paths
-    };
+    return documentForVersion;
   }
 
   public static async loadPluginMetadata(


lib/utils/rewrite-nullable-for-oas31.util.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  if (obj.nullable !== true) {
+    return;
+  }
+  delete obj.nullable;
+
+  if (typeof obj.type === 'string') {
+    obj.type = [obj.type, 'null'];
+    return;
+  }
```
simonboeuf-kpler, line 29: We do not yet cover a typed `$ref`/`allOf` composite (the shape the factory emits for `@ApiProperty({ type: Dto, nullable: true })`). It becomes `{ type: ['object','null'], allOf: [...] }`, which can never accept null: the type allows null, but the allOf still forbids it.

The fix guards both type branches so composites fall through to the `oneOf` wrap you already have. The same block also corrects `type: 'null'`, which currently produces the invalid `['null','null']`. It replaces the two type checks (lines 26-35):

```ts
  const isComposite =
    '$ref' in obj || 'allOf' in obj || 'anyOf' in obj || 'oneOf' in obj;

  if (typeof obj.type === 'string' && !isComposite) {
    obj.type = obj.type === 'null' ? 'null' : [obj.type, 'null'];
    return;
  }
  if (Array.isArray(obj.type) && !isComposite) {
    if (!obj.type.includes('null')) {
      obj.type = [...obj.type, 'null'];
    }
    return;
  }
```

Failing tests:

```typescript
it('wraps a typed nullable $ref/allOf composite in oneOf', () => {
  const schema: any = { type: 'object', nullable: true, allOf: [{ $ref: '#/components/schemas/Profile' }] };
  rewriteNullableForOas31(schema);
  expect(schema).toEqual({ oneOf: [{ type: 'object', allOf: [{ $ref: '#/components/schemas/Profile' }] }, { type: 'null' }] });
});

it('does not produce a ["null","null"] union', () => {
  const schema: any = { type: 'null', nullable: true };
  rewriteNullableForOas31(schema);
  expect(schema).toEqual({ type: 'null' });
});
```

lib/utils/rewrite-nullable-for-oas31.util.ts
```diff
@@ -0,0 +1,42 @@
...
+  const obj = value as Record<string, unknown>;
+  for (const nested of Object.values(obj)) {
+    rewriteNullableForOas31(nested);
+  }
+
+  if (obj.nullable !== true) {
+    return;
+  }
+  delete obj.nullable;
+
```
simonboeuf-kpler, line 25: We should handle both `enum` and `const`. Failing tests example: 

```typescript
it('adds null to enum values', () => {
  const schema: any = { type: 'string', enum: ['a', 'b'], nullable: true };
  rewriteNullableForOas31(schema);
  expect(schema).toEqual({ type: ['string', 'null'], enum: ['a', 'b', null] });
});
it('keeps a nullable const satisfiable', () => {
  const schema: any = { type: 'string', const: 'x', nullable: true };
  rewriteNullableForOas31(schema);
  expect(schema).toEqual({ type: ['string', 'null'], enum: ['x', null] });
});
```



```suggestion
 if (Array.isArray(obj.enum)) {
    if (!obj.enum.includes(null)) {
      obj.enum = [...obj.enum, null];
    }
  } else if ('const' in obj) {
    obj.enum = [obj.const, null];
    delete obj.const;
  }
```

lib/utils/rewrite-nullable-for-oas31.util.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (typeof obj.type === 'string') {
+    obj.type = [obj.type, 'null'];
```
simonboeuf-kpler: For type: `'null'` this produces `['null','null']`, which is not valid 

Failing test example:

```
it('does not produce a ["null","null"] union', () => {
  const schema: any = { type: 'null', nullable: true };
  rewriteNullableForOas31(schema);
  expect(schema).toEqual({ type: 'null' });
});
```


```suggestion
    obj.type = obj.type === 'null' ? 'null' : [obj.type, 'null'];
```

lib/utils/rewrite-nullable-for-oas31.util.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  const obj = value as Record<string, unknown>;
+  for (const nested of Object.values(obj)) {
+    rewriteNullableForOas31(nested);
+  }
+
+  if (obj.nullable !== true) {
+    return;
+  }
```
simonboeuf-kpler: This leaves `nullable: false` in the document. Acceptable but we can easily drop it here. Failing test: 

```typescript
it('drops a redundant nullable: false', () => {
  const schema: any = { type: 'string', nullable: false };
  rewriteNullableForOas31(schema);
  expect(schema).toEqual({ type: 'string' });
});
```


```suggestion
  if (obj.nullable !== true) {
    delete obj.nullable;
    return;
  }
```

lib/utils/rewrite-nullable-for-oas31.util.ts
```diff
@@ -0,0 +1,42 @@
...
+    obj.type = [obj.type, 'null'];
+    return;
+  }
+  if (Array.isArray(obj.type)) {
+    if (!obj.type.includes('null')) {
+      obj.type = [...obj.type, 'null'];
+    }
+    return;
+  }
+
```
simonboeuf-kpler, line 36: When a `oneOf/anyOf` already has a null branch, the wrap below nests it a second time. This guard skips the redundant wrap. It needs one small helper at module scope.

Failing test example: 

```typescript
it('drops a redundant nullable when the union already admits null', () => {
  const schema: any = { nullable: true, oneOf: [{ type: 'string' }, { type: 'null' }] };
  rewriteNullableForOas31(schema);
  expect(schema).toEqual({ oneOf: [{ type: 'string' }, { type: 'null' }] });
});
```
 
Suggested fix

```
  const branches = [obj.oneOf, obj.anyOf].find(Array.isArray) as unknown[] | undefined;
  if (branches?.some(admitsNull)) {
    return;
  }
  
// helper function, probably at module scope
function admitsNull(schema: unknown): boolean {
  if (!schema || typeof schema !== 'object') {
    return false;
  }
  const s = schema as Record<string, unknown>;
  return s.type === 'null' || (Array.isArray(s.type) && s.type.includes('null'));
}
```

test/utils/rewrite-nullable-for-oas31.util.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+  });
+
+  it('leaves 3.0-compatible schemas without nullable unchanged', () => {
+    const schema: any = { type: 'string' };
+
+    rewriteNullableForOas31(schema);
+
+    expect(schema).toEqual({ type: 'string' });
+  });
```
simonboeuf-kpler, line 63: This recursion goes into every value, including `example`, `examples`, `default`, `const` payloads, and `x-*` extensions. A correct fix needs position-aware traversal, which is a larger change, just flagging this as a failing test here, let me know if you would like a fix suggestion

```suggestion
  });
  
  it('does not rewrite a schema-shaped example payload', () => {
  const schema: any = { type: 'object', properties: { name: { type: 'string' } }, example: { note: { type: 'string', nullable: true } } };
  rewriteNullableForOas31(schema);
  expect(schema.example).toEqual({ note: { type: 'string', nullable: true } });
});
```

lib/utils/rewrite-nullable-for-oas31.util.ts
```diff
@@ -0,0 +1,42 @@
...
+  const obj = value as Record<string, unknown>;
+  for (const nested of Object.values(obj)) {
+    rewriteNullableForOas31(nested);
+  }
+
+  if (obj.nullable !== true) {
+    return;
+  }
+  delete obj.nullable;
```
simonboeuf-kpler: This leaves `nullable: false` in the document. Acceptable but we can easily drop it here. Failing test: 

```typescript
it('drops a redundant nullable: false', () => {
  const schema: any = { type: 'string', nullable: false };
  rewriteNullableForOas31(schema);
  expect(schema).toEqual({ type: 'string' });
});
```


```suggestion
  if (obj.nullable !== true) {
    delete obj.nullable;
    return;
  }
```

lib/utils/rewrite-nullable-for-oas31.util.ts
simonboeuf-kpler, line 23: This leaves `nullable: false` in the document. Acceptable but we can easily drop it here. Failing test: 

```typescript
it('drops a redundant nullable: false', () => {
  const schema: any = { type: 'string', nullable: false };
  rewriteNullableForOas31(schema);
  expect(schema).toEqual({ type: 'string' });
});
```


```suggestion
  if (obj.nullable !== true) {
    delete obj.nullable;
    return;
  }
```

simonboeuf-kpler (commented):
Thanks for the PR! 

We run our own 3.1 nullable normalization so we have hit several edge cases in real specs. I added an inline comment for each one, and every comment has a failing test and a suggested fix.

The last comment, on the spec file, is a heads-up only. A correct fix needs position-aware traversal, which is a larger change. I would be glad to help with that if it is useful.

micalevisk:
Thanks for taking this on, and thanks @simonboeuf-kpler for the detailed inline review.

Overall: **right approach, under-built.** A post-generation document walker is the correct shape for this fix — it catches everything, including the `nullable: true` that the CLI plugin injects, which a factory-level fix would miss. But the current implementation has correctness bugs on the exact shapes this library emits, plus one design flaw that can't be patched incrementally.

## On the existing inline comments

All five are legitimate. Ranked by how much they should block:

| # | Comment | Severity |
|---|---|---|
| 1 | Typed `$ref`/`allOf` composite → `{type:['object','null'], allOf:[…]}` | **Blocking** |
| 2 | Blind recursion into `example` / `default` / `x-*` | **Blocking (design)** |
| 3 | `enum` / `const` not extended with `null` | High |
| 4 | Double-wrap when `oneOf` already has a null branch | Medium |
| 5 | `nullable: false` left in the document | Low |

**`#1` is the headline bug, and it's provable — not hypothetical.** `lib/services/schema-object-factory.ts#L649-L656` emits exactly that shape for `@ApiProperty({ type: SomeDto, nullable: true })`:

```ts
...validMetadataObject,
...(validMetadataObject['nullable'] ? { type: 'object' } : {}),
allOf: [{ $ref }]
```

The `typeof obj.type === 'string'` branch fires first, producing `{ type: ['object','null'], allOf: [{ $ref }] }`. That is *unsatisfiable for null*: the type union admits `null`, but the `allOf` branch still requires the referenced object. So the single most common nullable-DTO case ends up broken in a way that's arguably worse than the status quo. The `isComposite` guard suggested above is the right fix.

**`#2` needs a structural fix, not a guard.** `rewriteNullableForOas31` recurses into every object value with no notion of *where* it is in the document. Any `example`, `examples`, `default`, or `x-*` extension whose payload happens to contain a `nullable: true` key gets rewritten as if it were a schema — a user documenting their own OpenAPI-ish config object would get their example silently corrupted. The fix is a position-aware walk: `paths` → path items → operations → parameters / requestBody / responses → content → `schema`, plus `components.*` and `webhooks`, applying the conversion only at genuine schema positions and recursing into `properties`, `items`, `additionalProperties`, `prefixItems`, `not`, and the combinators. That's ~150 lines instead of 42, but it's the only version that's actually safe.

## A few things not yet raised

**Annotation keywords get swallowed by the `oneOf` wrap.** The catch-all branch moves *every* sibling key inside the union, so `{ nullable: true, description: 'Owner', allOf: [{ $ref }] }` becomes `{ oneOf: [{ description: 'Owner', allOf: […] }, { type: 'null' }] }`. `description`, `title`, `deprecated`, and `readOnly` still apply at the union level and should stay outside it — otherwise tooling that reads the property description loses it.

**The wrap fires on schemas that don't need it.** `{ nullable: true, description: 'x' }` — no `type`, no `$ref` — becomes `{ oneOf: [{ description: 'x' }, { type: 'null' }] }`. A typeless schema already admits `null`; the correct action there is to drop `nullable` and return.

**In-place mutation of possibly shared objects.** The walker mutates the document deeply, but `documentForVersion` is only a *shallow* spread of `baseDocument`, and the factory builds schemas with `omit(...)`, which is also shallow — so nested objects such as `metadata.items` and `metadata.enumSchema` may still be the same references held in the decorator metadata. An app that calls `createDocument` twice (a 3.1 doc and a 3.0 doc, or one document per API version — not exotic) risks the 3.1 rewrite leaking into the 3.0 document. Worth a regression test: build a 3.1 document, then a 3.0 document from the same controllers, and assert the 3.0 one still has `nullable: true`.

**Nothing tests the version gate.** All five tests call the util directly. Nothing asserts that `createDocument` applies it for `3.1.0` and skips it for `3.0.0` — which is the actual contract of this fix. @simonboeuf-kpler offered e2e tests in the issue; an e2e case that generates both a 3.0 and a 3.1 document from the same fixture app would cover both this and the shared-mutation point above.

## Out of scope here, but adjacent

The reverse direction is still broken and this PR shouldn't try to fix it — flagging it so it isn't assumed solved once this merges. Since #3897 the response object factory emits `oneOf: [<schema>, { type: 'null' }]` **unconditionally**, so **3.0 documents get 3.1 syntax** in inline responses. That means the "one document mixes both dialects" complaint in #4063 is only half-addressed by this PR: it fixes 3.1 documents carrying 3.0 spelling, but 3.0 documents carrying 3.1 spelling remains. The proper fix is to thread the declared OpenAPI version down through `SwaggerScanner` → `SwaggerExplorer` → `ResponseObjectFactory` → `ResponseObjectMapper` so the response path can pick the correct spelling per version, defaulting to `3.0.0` for backward compatibility. Separate PR.

## What I'd like to see before merge

- #1 and #3 fixed (both correctness)
- #4 and #5 taken as suggested (cheap)
- #2 restructured into a position-aware walker
- an e2e test generating both a 3.0 and a 3.1 document from the same fixture app

One offer to save you the work on `#2`: I have a working position-aware walker locally that already handles the composite case, `enum` null-extension, annotation preservation, and unwinding the synthetic `{ type: 'object', allOf: [{ $ref }] }` wrapper. Happy to hand it over so you can fold it into this PR rather than rebuilding it — just say the word.

kamilmysliwiec:
ref https://github.com/nestjs/swagger/pull/4070

PR #4070, opened 2026-08-23
fix: spell nullability per the declared OpenAPI version

### LINKED ISSUE #4063: Schema properties keep the 3.0 nullable keyword in documents declared as OpenAPI 3.1

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: closes #4063

A document can currently mix both OpenAPI dialects, in **both** directions, and the declared version changes neither output:

1. **3.1 documents carry 3.0 spelling.** `DocumentBuilder.setOpenAPIVersion('3.1.0')` sets `openapi: 3.1.0`, but the schema object factory still emits `{ "type": "string", "nullable": true }`. 3.1 is JSON Schema 2020-12, which removed `nullable` and ignores unknown keywords, so a strict 3.1 consumer reads that as a plain non-nullable string. Nullability is lost silently — no tool warns, and generated clients break on the first real `null`.

2. **3.0 documents carry 3.1 spelling.** Since #3897 the response object factory emits `oneOf: [<schema>, { "type": "null" }]` **unconditionally**. `type: 'null'` is not valid in 3.0, so 3.0 documents have been emitting 3.1-only syntax for inline nullable responses.

## What is the new behavior?

The emission follows the declared document version, in both directions, through a single normalization step rather than version branching at each emission site.

**The factories always emit the 3.0 spelling.** `SwaggerModule.createDocument` then normalizes the finished document via a new `convertNullableToOas31` walker when it declares 3.1 or later:

| Input (3.0 spelling) | Output (3.1 spelling) |
|---|---|
| `{ type: 'string', nullable: true }` | `{ type: ['string', 'null'] }` |
| `{ type: ['string','number'], nullable: true }` | `{ type: ['string','number','null'] }` |
| `{ type: 'string', enum: ['a','b'], nullable: true }` | `{ type: ['string','null'], enum: ['a','b',null] }` |
| `{ $ref, nullable: true }` | `{ anyOf: [{ $ref }, { type: 'null' }] }` |
| `{ type: 'object', nullable: true, allOf: [{ $ref }] }` | `{ anyOf: [{ $ref }, { type: 'null' }] }` |
| `{ nullable: false }` | `{}` |

**3.0.x documents** keep `nullable: true`, and the response path now produces the 3.0 spelling instead of the `type: 'null'` it has been emitting since #3897. Because the internals speak one dialect and only the document boundary translates, this needs no version plumbing through the scanner, explorer or factories — `createDocument` is the only place that reads `config.openapi`.

Nullable **responses** in 3.1 documents are consequently spelled as a type union or `anyOf` rather than `oneOf`, which matches how nullable *properties* are emitted and makes the document internally consistent:

| `@ApiResponse` | 3.0 output | 3.1 output |
|---|---|---|
| `{ type: CatDto, nullable: true }` | `{ nullable: true, type: 'object', allOf: [{ $ref }] }` | `{ anyOf: [{ $ref }, { type: 'null' }] }` |
| `{ type: CatDto, isArray: true, nullable: true }` | `{ type: 'array', items: { $ref }, nullable: true }` | `{ type: ['array','null'], items: { $ref } }` |
| `{ type: String, nullable: true }` | `{ type: 'string', nullable: true }` | `{ type: ['string','null'] }` |

Because the walker runs on the finished document, it also normalizes `nullable: true` injected by the **CLI plugin** (`isNullableUnion` in `model-class.visitor.ts`) — the case the issue calls out as having no decorator-level workaround.

### Relationship to #4064

This is an alternative to @official-burak's #4064, which takes the same document-walker approach — credit to that PR for the shape of the fix, and to @simonboeuf-kpler for the review there that mapped out the edge cases. I had this branch in progress when that PR landed; opening it since it covers the cases raised in that review plus the 3.0 direction. Happy to close this in favour of #4064 if the author prefers to fold these in, or to have the two merged — whatever the maintainers prefer.

A correction to an earlier revision of this description, since it argued the opposite: this walker **also** recurses generically, the way #4064 does. I first wrote it as a structural traversal that descended only through known OpenAPI keys, on the theory that converting solely at genuine schema positions was safer. That turned out to be the weaker design — an allowlist can only ever cover the keywords its author thought of, and a differential test against the generic version showed the structural one silently skipped `$defs`, `if`/`then`/`else`, `contains`, `propertyNames`, `unevaluatedItems`, `unevaluatedProperties`, `dependentSchemas`, `contentSchema` and tuple-form `items`. #4064's instinct was right; the last commit here adopts it and drops ~100 lines.

What this walker adds on top is a **closed blocklist**: it skips the keys that hold free-form user data — `example`, `examples`, `default`, `const`, `enum` — plus `x-` extensions. The OpenAPI object model has no other free-form positions, so unlike an allowlist that blocklist is provably complete, and a `nullable: true` sitting inside an example or a vendor extension is left alone. Those key rules are not applied inside name-keyed maps (`properties`, `patternProperties`, `$defs`, `schemas`, `headers`), so a header called `x-total` or a property called `example` is still converted.

The remaining differences from #4064 as it stands:

- **Typed composites.** `@ApiProperty({ type: SomeDto, nullable: true })` emits `{ type: 'object', nullable: true, allOf: [{ $ref }] }` (`schema-object-factory.ts#L649-L656`). Converting only the `type` there yields `{ type: ['object','null'], allOf: [{ $ref }] }`, which can never accept `null` — the union admits it, the `allOf` still forbids it. This unwinds the synthetic wrapper into `anyOf: [{ $ref }, { type: 'null' }]`.
- **`enum` gets `null` appended**, as the issue's expected behaviour specifies.
- **Annotation keywords stay outside the union.** `description`, `title`, `deprecated`, and `readOnly` still apply at the union level, so they are kept as siblings of `anyOf` rather than swallowed into its first branch.
- **`nullable: false` is dropped** rather than left in the 3.1 output.
- **Idempotent** — re-running over a converted document is a no-op, and an existing `null` branch is never duplicated.
- **The 3.0 direction** (point 2 above), which #4064 does not cover.

`anyOf` is used rather than `oneOf` for the reference case: with `oneOf`, a `$ref` whose target itself admits `null` would match two branches and fail validation.

- [ ] Yes
- [x] No

3.1 documents change only to match the 3.1 schema language. 3.0 documents change only for inline nullable responses, which were emitting invalid-for-3.0 `type: 'null'` since #3897; that output is now valid 3.0.

## Other information

**Scope** — 5 files under `lib/`, of which the walker (`lib/utils/nullable-to-oas31.util.ts`, 134 lines) is the only new one. `ResponseObjectMapper` and `ResponseObjectFactory` change by a few lines each, `SwaggerModule` calls the walker, and `SchemaObject.type` widens to `string | string[]` for 3.1 type unions.

**Tests** — 379 lines of unit tests for the walker (`test/utils/nullable-to-oas31.spec.ts`) covering every row of the tables above plus traversal positions, non-schema payloads, and idempotency; `ResponseObjectMapper` / `ResponseObjectFactory` specs assert the 3.0 spelling at that layer, with the 3.1 outcome covered by normalizing a document end to end; and e2e coverage in `validate-schema.e2e-spec.ts` that generates the same fixture app as both a 3.0 and a 3.1 document and validates each against its own dialect. New nullable primitive and enum fixtures were added to `CreateCatDto` to exercise this. Full suite: 496 unit + 133 e2e passing, clean build, no new lint warnings.

**Note on the `e2e/api-spec.json` diff** — most of it is key reordering unrelated to this change. `master`'s committed spec is stale with respect to its own `zod` version: checking out `master` and regenerating the file produces 124 lines of the same churn with no source changes at all. The only substantive additions here are the new `nullablePrimitive` / `nullableEnum` fixture properties. Happy to split that regeneration into its own commit or PR if that makes review easier.

**Reviewing the last commit** — the third commit is a refactor of the two fixes before it, so `master...HEAD` is the diff worth reading; the intermediate version-threading state has no independent value and these three squash cleanly if that is preferred.

---

Thanks to Claude Opus 5 + Fable

### DIFF
--- a/e2e/api-spec.json
+++ b/e2e/api-spec.json
@@ -768,10 +768,10 @@
                 "type": "object",
                 "properties": {
                   "name": {
-                    "description": "Cat name from Zod",
-                    "example": "Milo",
                     "type": "string",
-                    "minLength": 1
+                    "minLength": 1,
+                    "description": "Cat name from Zod",
+                    "example": "Milo"
                   },
                   "age": {
                     "type": "integer",
@@ -866,18 +866,16 @@
                 "type": "object",
                 "properties": {
                   "species": {
-                    "description": "Species enum from Zod",
-                    "title": "Species",
-                    "example": "cat",
                     "type": "string",
                     "enum": [
                       "cat",
                       "dog"
-                    ]
+                    ],
+                    "description": "Species enum from Zod",
+                    "title": "Species",
+                    "example": "cat"
                   },
                   "contact": {
-                    "description": "Preferred contact from Zod",
-                    "title": "PreferredContact",
                     "anyOf": [
                       {
                         "type": "string",
@@ -888,9 +886,9 @@
                         "type": "object",
                         "properties": {
                           "phone": {
+                            "type": "string",
                             "description": "Phone number from Zod",
-                            "example": "123-456",
-                            "type": "string"
+                            "example": "123-456"
                           }
                         },
                         "required": [
@@ -898,65 +896,67 @@
                         ]
                       }
                     ],
+                    "description": "Preferred contact from Zod",
+                    "title": "PreferredContact",
                     "example": "owner@example.com"
                   },
                   "profile": {
-                    "title": "CatProfile",
-                    "description": "Nested cat profile from Zod",
                     "type": "object",
                     "properties": {
                       "nickname": {
+                        "type": "string",
                         "description": "Nested nickname from Zod",
-                        "example": "Captain Whiskers",
-                        "type": "string"
+                        "example": "Captain Whiskers"
                       },
                       "temperament": {
-                        "description": "Nested temperament from Zod",
-                        "example": "playful",
                         "type": "string",
                         "enum": [
                           "playful",
                           "calm"
-                        ]
+                        ],
+                        "description": "Nested temperament from Zod",
+                        "example": "playful"
                       },
                       "notes": {
+                        "default": "Indoor only",
                         "description": "Nested notes from Zod",
                         "example": "Indoor only",
                         "deprecated": true,
-                        "default": "Indoor only",
                         "type": "string"
                       },
                       "traits": {
-                        "description": "Nested trait list from Zod",
                         "type": "array",
                         "items": {
                           "type": "object",
                           "properties": {
                             "label": {
+                              "type": "string",
                               "description": "Trait label from Zod",
-                              "example": "affectionate",
-                              "type": "string"
+                              "example": "affectionate"
                             },
                             "score": {
-                              "description": "Trait score from Zod",
-                              "example": 8,
                               "type": "number",
                               "minimum": 0,
-                              "maximum": 10
+                              "maximum": 10,
+                              "description": "Trait score from Zod",
+                              "example": 8
                             }
                           },
                           "required": [
                             "label",
                             "score"
                           ]
-                        }
+                        },
+                        "description": "Nested trait list from Zod"
                       }
                     },
                     "required": [
                       "nickname",
                       "temperament",
                       "traits"
-                    ]
+                    ],
+                    "title": "CatProfile",
+                    "description": "Nested cat profile from Zod"
                   }
                 },
                 "required": [
@@ -1038,9 +1038,9 @@
                 "type": "object",
                 "properties": {
                   "override": {
+                    "type": "string",
                     "description": "Body override from Zod",
-                    "example": "zod-body",
-                    "type": "string"
+                    "example": "zod-body"
                   },
                   "count": {
                     "type": "integer",
@@ -1639,11 +1639,11 @@
             "required": true,
             "in": "query",
             "schema": {
-              "description": "Query override from Zod",
-              "example": 3,
               "type": "integer",
               "minimum": 1,
-              "maximum": 9007199254740991
+              "maximum": 9007199254740991,
+              "description": "Query override from Zod",
+              "example": 3
             }
           },
           {
@@ -1651,9 +1651,9 @@
             "required": true,
             "in": "query",
             "schema": {
+              "type": "boolean",
               "description": "Boolean flag from Zod",
-              "example": true,
-              "type": "boolean"
+              "example": true
             }
           },
           {
@@ -2108,10 +2108,10 @@
             "required": true,
             "in": "path",
             "schema": {
-              "description": "Cat identifier from Zod",
               "type": "string",
               "minLength": 3,
-              "pattern": "^cat_[0-9]+$"
+              "pattern": "^cat_[0-9]+$",
+              "description": "Cat identifier from Zod"
             }
           },
           {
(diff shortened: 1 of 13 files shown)


kamilmysliwiec:
lgtm

PR #4080, opened 2026-08-28
fix(plugin): use namespace imports for ESM metadata type references

### LINKED ISSUE #4079: V12: plugin generates invalid code under ESM output
## CLAUDE generated 🤖

`@nestjs/swagger`'s CLI plugin generates invalid code under ESM output. When a `@ApiProperty` type is an enum imported from a different file, the plugin rewrites the reference to `(await import("...")).EnumName` inside the generated `static _OPENAPI_METADATA_FACTORY()` method — but never marks that method `async`. The compiled file throws `SyntaxError: Unexpected reserved word` on load, crashing the whole app at boot.

### Minimal reproduction

https://github.com/Roman991/nestjs-swagger-esm-repro

```bash
npm install
npm run build
npm start

Root cause

- dist/plugin/utils/plugin-utils.js, replaceImportPath(), esmCompatible branch (~line 210-216): emits (${typeImportStatement}).${typeName}, where typeImportStatement is await import("...").
- dist/plugin/visitors/model-class.visitor.js, ~line 137: builds the method with:

factory.createMethodDeclaration(
  [factory.createModifier(ts.SyntaxKind.StaticKeyword)],
  ...
)

  — only static, regardless of whether the body ends up containing await.

esmCompatible is auto-detected from moduleResolution: "nodenext" + "type": "module" (utils/module-format.util.js), so any Nest 12 project doing the official ESM migration hits this as soon as it has one DTO/entity with a cross-file enum property.

Expected behavior

_OPENAPI_METADATA_FACTORY should be generated as static async _OPENAPI_METADATA_FACTORY() whenever its body ends up containing an await import(...).

Suggested fix

In model-class.visitor.js, add the AsyncKeyword modifier to the generated method whenever any property went through the options.esmCompatible branch of replaceImportPath (i.e. the body will contain await):

const modifiers = [factory.createModifier(ts.SyntaxKind.StaticKeyword)];
if (needsAwait) {
  modifiers.push(factory.createModifier(ts.SyntaxKind.AsyncKeyword));
}
const method = factory.createMethodDeclaration(
  modifiers,
  undefined,
  factory.createIdentifier(METADATA_FACTORY_NAME),
  undefined,
  undefined,
  [],
  undefined,
  factory.createBlock([factory.createReturnStatement(returnValue)], true),
);

needsAwait just needs to be tracked while building returnValue — replaceImportPath already knows this internally, it just isn't propagated back up to the method-declaration call site.

Environment

- @nestjs/swagger: 12.0.0
- @nestjs/cli: 12.0.0
- typescript: 6.0.3
- Node.js: v26.7.0

### PR DESCRIPTION
**Pull request by lazerg**
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: closes #4079

When the plugin emits ESM, a property whose type lives in another file is rewritten to `(await import("./role.enum.js")).Role`, and that expression lands inside the synchronous `static _OPENAPI_METADATA_FACTORY()`:

```js
    return { role: { required: true, enum: (await import("./role.enum.js")).Role } };
}
```

Node throws `SyntaxError: Unexpected reserved word` when the file loads, so the app never boots.

That `(await import(...))` form comes from `convertToAsyncImport()`, which readonly mode needs because the generated `metadata.ts` awaits its imports inside an async function. Transform mode has no async context to place it in. `_OPENAPI_METADATA_FACTORY` is called synchronously by `ModelPropertiesAccessor`, the explorers and the type helpers, so marking the method `async` would hand all of them a promise in place of the metadata object.

## What is the new behavior?

In ESM output each referenced module is hoisted to a namespace import at the top of the file, and the metadata reads the binding off it, which keeps the lookup synchronous:

```js
import * as openapi_import_0 from "./role.enum.js";
...
        return { role: { required: true, enum: openapi_import_0.Role } };
    }
```

That is the ESM equivalent of the `require("./role.enum")` the CommonJS output already emits. It also covers the same expression in generated controller decorators, which sit outside the factory method and so would not have been helped by an `async` modifier.

- [ ] Yes
- [x] No

## Other information

A new case in `model-class-visitor.spec.ts` builds a real `ts.Program` over a two-file fixture and asserts the emitted DTO carries the namespace import and no `await`. The `parameter-property.dto` fixture had the broken output baked into its expectation, so it is updated to match.

### DIFF
--- a/lib/plugin/utils/plugin-utils.ts
+++ b/lib/plugin/utils/plugin-utils.ts
@@ -342,6 +342,7 @@ export function replaceImportPath(
         convertToAsyncImport(typeReference);
       return {
         typeReference: `(${typeImportStatement}).${typeName}`,
+        typeName,
         importPath: relativePath
       };
     }
--- a/lib/plugin/utils/type-reference-to-identifier.util.ts
+++ b/lib/plugin/utils/type-reference-to-identifier.util.ts
@@ -77,9 +77,8 @@ export function typeReferenceToIdentifier(
     importPath &&
     typeName
   ) {
-    // ESM has no synchronous `require`, and the generated metadata is read
-    // synchronously, so the referenced module is hoisted to a namespace import
-    // instead of being awaited inline.
+    // The metadata is read synchronously, so the module is hoisted to a
+    // namespace import rather than awaited inline.
     const namespace = registerEsmTypeImport(importPath, typeImports);
 
     let ref = `${namespace}.${typeName}`;
@@ -100,7 +99,7 @@ export function typeReferenceToIdentifier(
 function registerEsmTypeImport(
   importPath: string,
   typeImports: Record<string, string>
-) {
+): string {
   if (!typeImports[importPath]) {
     const index = Object.keys(typeImports).length;
     typeImports[importPath] = `${OPENAPI_NAMESPACE}_import_${index}`;
--- a/lib/plugin/utils/type-reference-to-identifier.util.ts
+++ b/lib/plugin/utils/type-reference-to-identifier.util.ts
@@ -1,6 +1,7 @@
 import { posix } from 'path';
 import * as ts from 'typescript';
 import { PluginOptions } from '../merge-options.js';
+import { OPENAPI_NAMESPACE } from '../plugin-constants.js';
 import { pluginDebugLogger } from '../plugin-debug-logger.js';
 import {
   convertPath,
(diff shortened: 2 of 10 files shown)


kamilmysliwiec:
already fixed here https://github.com/nestjs/swagger/pull/4082 but thanks anyway!

Issue #4088, opened 2026-09-01
Standard Schema query params are duplicated when combined with @ApiQuery


TadejPolajnar:
## Is there an existing issue for this?

- [x] I have searched the existing issues

## Current behavior

When a handler combines a Standard Schema query parameter (`@Query({ schema })`) with an `@ApiQuery()` for the same field, the generated document contains **two entries** for that parameter instead of one:

```json
[
  {
    "name": "limit", "in": "query", "required": false,
    "schema": { "type": "integer", "minimum": 1, "maximum": 100 }
  },
  {
    "name": "limit", "in": "query", "required": false,
    "description": "Max results",
    "schema": { "type": "number" }
  }
]
```

Neither entry is complete — the Standard Schema one has the accurate schema but no `description`; the `@ApiQuery` one has the description but a weaker schema. This is invalid per OpenAPI 3.x, which requires parameters to be unique by `name` + `in`.

Dropping `@ApiQuery` removes the duplicate but leaves the parameter undescribed: Zod's `.describe()` lands on `schema.description`, not the parameter-level `description`. So there is currently no way to get one parameter with both an accurate schema and a description.

## Minimum reproduction code

```js
// repro.mjs — decorators applied programmatically so it runs on plain Node
import 'reflect-metadata';
import { Controller, Get, Module, Query } from '@nestjs/common';
import { NestFactory } from '@nestjs/core';
import { ApiQuery, DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
import { createSchema } from 'zod-openapi';
import { z } from 'zod';

const convert = (schema, { schemaType }) => {
  if (!(schema instanceof z.ZodType)) return undefined;
  const c = createSchema(schema, { io: schemaType, openapiVersion: '3.0.0' });
  return { schema: c.schema, components: c.components };
};

class ItemsController {
  list(q) { return [q]; }
}
const descriptor = Object.getOwnPropertyDescriptor(ItemsController.prototype, 'list');

Query({ schema: z.strictObject({ limit: z.coerce.number().int().min(1).max(100).optional() }) })(
  ItemsController.prototype, 'list', 0
);
// tsc/SWC emit this; required or the explorer skips the parameter entirely.
Reflect.defineMetadata('design:paramtypes', [Object], ItemsController.prototype, 'list');
Get()(ItemsController.prototype, 'list', descriptor);
ApiQuery({ name: 'limit', required: false, type: Number, description: 'Max results' })(
[...]

PR #4090, opened 2026-09-02
fix(swagger): merge standard schema query params with @ApiQuery

TadejPolajnar:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #4088

A handler that combines a Standard Schema query parameter with an `@ApiQuery()` for the same field emits **two** parameter entries instead of one merged entry:

```ts
@Get()
@ApiQuery({ name: 'limit', required: false, type: Number, description: 'Max results' })
findAll(
  @Query({ schema: z.strictObject({ limit: z.coerce.number().int().min(1).max(100).optional() }) })
  query: unknown
) {}
```

```jsonc
// two "limit" entries, neither complete
[
  { "name": "limit", "in": "query",
    "schema": { "type": "integer", "minimum": 1, "maximum": 100 } },  // no description
  { "name": "limit", "in": "query", "description": "Max results",
    "schema": { "type": "number" } }                                  // weaker schema
]
```

This is invalid per OpenAPI 3.x (parameters must be unique by `name` + `in`), and there was no way to get one parameter with both an accurate schema and a description — Zod's `.describe()` lands on `schema.description`, not the parameter-level `description`.

**Why it happens.** `exploreApiParametersMetadata` merges reflected and explicit parameters using `hasSameParameterIdentity`:

```ts
// lib/explorers/api-parameters.explorer.ts
return left.name === right.name && left.in === right.in;
```

At that point a Standard Schema query param is still a single **unnamed** object carrying `standardSchema` — the fan-out into one param per property happens later, in `SchemaObjectFactory.createSchemaOverrideParam`. So the comparison is `undefined === 'limit'` → `false`, `unionWith` keeps both, and the later fan-out turns the unnamed entry into a second named `limit`.

Class DTOs are immune because `ParametersMetadataMapper.transformModelToProperties` expands them into named properties *before* the merge.

## What is the new behavior?

Standard Schema params are expanded into named params **before** the identity merge, so the existing merge machinery handles them exactly as it already handles class DTOs.

```jsonc
// one entry, accurate schema + description
[
  { "name": "limit", "in": "query", "description": "Max results",
    "schema": { "type": "integer", "minimum": 1, "maximum": 100 } }
]
```

The fix, in `lib/explorers/api-parameters.explorer.ts`:

```ts
-  let properties = reflectedParametersAsProperties;
+  let properties = reflectedParametersAsProperties.flatMap((param) => {
+    const expanded = schemaObjectFactory.expandStandardSchemaParam(param, schemas);
+    return expanded ?? [param];
+  });
```

plus a public `expandStandardSchemaParam` on `SchemaObjectFactory` that converts the schema and returns one param per property (or `undefined` when the converted schema is not a plain object).

No new conversion logic — it reuses `getSchemaOverride` and `createSchemaOverrideParam` unchanged, including their existing `type === 'object' && 'properties' in schema` guard. Non-object schemas, body params and `@Query('name', { schema })` hit that method's non-array branches and keep their current representation.

Expanding in `ParametersMetadataMapper` (where class DTOs expand) would need the user-supplied `standardSchemaConverter` — but that lives on `SchemaObjectFactory`, which `SwaggerScanner` re-instantiates per document, while the mapper is a module-scope singleton. The explorer already receives the per-document factory (cab7b984) and already branches on `standardSchema`.

Adds `test/explorers/api-parameters.explorer.spec.ts` (6 tests). `e2e/api-spec.json` updates for the existing `standard-query-conflict` fixture, which had the duplicate committed into it. Suite green: 504 unit, 133 e2e.

### Two calls worth your eye

**1. `required` on a conflict.** That fixture pairs a Zod schema where `filter` is required with `@ApiQuery({ required: false })`. The explicit decorator now wins, so the fixture changes `"required": true` → `false` — consistent with class DTOs, but a semantic choice. Happy to invert it.

**2. It also fixes an unreported case.** Two `@Query({ schema })` params on one handler currently collide (`undefined === undefined`) and one is silently dropped. After expansion both survive. Untested here as it's outside #4088 — glad to add one.

Unrelated follow-up: the e2e helper at `e2e/standard-schema.e2e-spec.ts:233` uses `.find()`, which is blind to duplicates — why this passed CI with the duplicate in the fixture.

- [ ] Yes
- [x] No

Documents that previously contained duplicate parameter entries will now contain one merged entry. That output was invalid per the OpenAPI spec, so this corrects it rather than changing a supported contract — but it does change generated output for affected routes.

## Other information

Verified against a codebase with 55 affected query routes (349 operations): all 55 resolve to single merged parameters.

One side effect worth flagging: where a schema and an `@ApiQuery` disagreed, the merged parameter keeps the schema-derived shape, so some generated types get *more* specific. A `z.union([z.boolean(), z.enum(['true','false'])])` param had been emitting just `boolean` (the weaker `@ApiQuery` entry won the downstream dedupe) and now emits both accepted forms.


lib/explorers/api-parameters.explorer.ts
```diff
@@ -53,7 +53,13 @@ export const exploreApiParametersMetadata = (
       parametersMetadata || {}
     );
 
-  let properties = reflectedParametersAsProperties;
+  let properties = reflectedParametersAsProperties.flatMap((param) => {
```
micalevisk, line 62: Right place for it. One thing to spell out in a short comment here: this has to run before `hasSameParameterIdentity` because Standard Schema params are still a single unnamed object at this point (unlike class DTOs, which `transformModelToProperties` already fanned out). Future readers will otherwise wonder why the explorer expands while `createFromModel` also does.

lib/services/schema-object-factory.ts
```diff
@@ -1083,6 +1083,23 @@ export class SchemaObjectFactory {
     return undefined;
   }
 
+  expandStandardSchemaParam(
+    param: ParamWithTypeMetadata,
+    schemas: Record<string, SchemaObject>
+  ): ParamWithTypeMetadata[] | undefined {
+    if (!param.standardSchema) {
```
micalevisk: This converts body params and named params (`@Query('x', { schema })`) only to discard the result, and `createFromModel` converts them again a few lines later. With one `@Body({ schema })` and one `@Query({ schema })` the user-supplied converter runs 3 times instead of 2. Besides the wasted work, a converter that isn't idempotent (e.g. one that mints unique component names) would register duplicates.

Please short-circuit before converting:

```suggestion
    if (!param.standardSchema || param.name || isBodyParameter(param)) {
```

Optional follow-on: after this PR the array-returning branch of `createSchemaOverrideParam` is only reachable from here, so you could move the `type === 'object' && 'properties' in schema` expansion into this method and leave `createSchemaOverrideParam` with the body/named cases only. That would also let you drop the `Array.isArray` + cast.

test/explorers/api-parameters.explorer.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+        ItemsController.prototype,
+        'list'
+      )!;
+
+      Query({
+        schema: z.strictObject({
+          limit: z.coerce.number().int().min(1).max(100).optional()
+        })
+      } as any)(ItemsController.prototype, 'list', 0);
+      Reflect.defineMetadata(
```
micalevisk: Please declare these controllers with real decorators like the rest of the suite does, e.g. `test/services/parameter-metadata-accessor.spec.ts` uses `@Query({ schema: querySchema }) query: unknown` inside a class and `design:paramtypes` is emitted fine under vitest. That removes the manual `Reflect.defineMetadata('design:paramtypes', ...)`, the `as any` casts and the `Object.getOwnPropertyDescriptor` dance in every test.

test/explorers/api-parameters.explorer.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+      ).toEqual(
+        expect.objectContaining({
+          required: true,
+          description: 'Search term',
+          schema: { type: 'string' }
+        })
+      );
+    });
+
+    it('should preserve schema-derived required flags when @ApiQuery omits them', () => {
```
micalevisk: The title doesn't match what's being tested. `@ApiQuery` defaults `required` to `true` (`defaultQueryOptions` in `api-query.decorator.ts`), so `required: true` on `mandatory` comes from the decorator default, not from the Zod schema. I checked on this branch: `@ApiQuery({ name: 'limit', description: 'x' })` on `z.number().optional()` produces `required: true`.

That's consistent with how class DTOs merge today (same `assign`), so I'm fine with the behavior, but the test should say what actually happens. Please rename this one and add a case for the optional field + `@ApiQuery` without `required`, asserting `required: true`, so the precedence is documented rather than implied.

test/explorers/api-parameters.explorer.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+          in: 'body',
+          type: Object,
+          required: true,
+          standardSchema: z.strictObject({ title: z.string() })
+        } as any,
+        {}
+      )).toBeUndefined();
+    });
+
+    it('should leave a named standard schema query param as a single parameter', () => {
```
micalevisk: Nit: this and the next test call `SchemaObjectFactory.expandStandardSchemaParam` directly, so they belong in `test/services/schema-object-factory.spec.ts`. Keep this spec about the explorer's output.

While you're there, please add the "two `@Query({ schema })` on one handler" case you mentioned in the description. I verified both params survive on this branch, and since the PR changes that behavior it should be pinned.

e2e/api-spec.json
```diff
@@ -1636,15 +1636,16 @@
           },
           {
             "name": "filter",
-            "required": true,
+            "required": false,
```
micalevisk, line 1639: Agree with `required: false` here: explicit `@ApiQuery` wins, same as class DTOs, and the same rule cab7b984 already asserts for the path and body conflict fixtures.

Note this file is *written* by `validate-schema.e2e-spec.ts`, not compared, so it isn't a regression guard on its own. Please add to `assertStandardSchemaDocument` an assertion that the `standard-query-conflict` parameters are unique by `(name, in)`, and that `filter` has `required === false` and the explicit description. Today `getParameter` uses `.find()`, which is why the duplicate sat in this fixture unnoticed.

micalevisk (changes requested):
Thanks for the thorough writeup and for tracking the root cause down to the ordering between `transformModelToProperties` and `hasSameParameterIdentity`. I agree with the approach: expanding Standard Schema params *before* the identity merge is the right place to fix this, and it keeps a single set of merge semantics (`mergeImplicitAndExplicit`) instead of adding a second dedupe path.

For context, #4110 is a competing fix for #4088 that dedupes after `mapParamTypes`; I requested changes there and asked for exactly what this PR does, so this is the one I want to move forward with.

I pulled the branch locally: 504 unit tests, 133 e2e, lint and build are all green. I also probed a few cases on top of the branch and found some things that need to change before merging. None of them are about the core approach.

### On the two calls you flagged

1. **`required` on a conflict: keep it as you have it (explicit `@ApiQuery` wins).** That's how class DTOs already behave, and the feature commit (cab7b984, #3915) encodes the same rule for the path and body conflict fixtures (`standard-param-conflict` asserts the explicit description on top of the Zod schema, `standard-body-conflict` asserts the explicit `required: false`). No need to invert it.

2. **Two `@Query({ schema })` on one handler:** yes please, add a unit test for it. I verified both survive on this branch (they used to collide on `undefined === undefined`), and since the PR changes that behavior it should be pinned by a test.

### Requested changes

- **Rewrite the new spec with decorator syntax.** The rest of the suite (e.g. `test/services/parameter-metadata-accessor.spec.ts`) declares controllers with real `@Query({ schema })` / `@Body({ schema })` decorators and `design:paramtypes` is emitted fine under vitest. Applying decorators programmatically plus a manual `Reflect.defineMetadata('design:paramtypes', ...)` and `as any` casts is a new pattern in this repo and makes the tests harder to read than they need to be. (inline)
- **Test 3 asserts the wrong thing.** `required: true` on `mandatory` comes from `@ApiQuery`'s default (`defaultQueryOptions.required = true`), not from the Zod schema. I verified that `@ApiQuery({ name: 'limit', description: '...' })` on `z.number().optional()` yields `required: true` on this branch. That's consistent with class DTOs (same `assign`), so I'm not asking you to change the behavior, but the test title claims something the code does not do. Please rename it, and add a case that documents the actual precedence (explicit default overrides the schema-derived `required`). (inline)
- **Add an e2e guard.** `e2e/api-spec.json` is written by `validate-schema.e2e-spec.ts`, not compared, and `getParameter` uses `.find()`, so today nothing at the e2e level would catch this bug coming back. Please add an assertion on the `standard-query-conflict` route that parameters are unique by `(name, in)`, plus `required === false` and the explicit description on `filter`. Same bar I asked for on #4110.
- **Avoid running the converter twice for body params.** `expandStandardSchemaParam` converts every `standardSchema` param, including body and named ones, only to throw the result away; `createFromModel` then converts them again. With one `@Body({ schema })` and one `@Query({ schema })` the user converter is invoked 3 times instead of 2. Please bail out on `param.name || isBodyParameter(param)` before calling `getSchemaOverride`. (inline, with an optional follow-on refactor)

### Nits

- Commit message: this repo uses `fix: ...` without a `(swagger)` scope (see `git log`).
- The last two tests exercise `SchemaObjectFactory.expandStandardSchemaParam` directly; they belong in `test/services/schema-object-factory.spec.ts` rather than the explorer spec.

### Gaps I'd take as follow-ups (not blocking)

- A Zod object with a registered id (`z.object({...}).meta({ id: 'PageQuery' })`) converts to a `$ref`, so it never hits the `type === 'object' && 'properties' in schema` branch, is never expanded, and combined with `@ApiQuery` ends up as a single param with `schema: {}`. That's broken on `master` too; fixing it means resolving the `$ref` against `schemas` before expanding.
- Explicit `type` on `@ApiQuery` is dropped in favor of the Zod schema (it's in `getSchemaOptionsKeys`, so `omitParamKeys` strips it), but explicit `enum` replaces the Zod schema entirely because `addEnumSchema` writes a `schema`. Worth documenting or aligning at some point.
- The `.find()` blindness in the e2e helper you mentioned.
- Bonus from this PR worth noting in the description: global parameters (`DocumentBuilder.addGlobalParameters`) with the same name as a Zod property now dedupe correctly too, since `unionWith` sees the expanded params.

kamilmysliwiec:
On the optional follow-on refactor (moving the `type === 'object' && 'properties' in schema` expansion into `expandStandardSchemaParam` and leaving `createSchemaOverrideParam` with the body/named cases) — I left that one out deliberately.

`createFromModel` is a public method on `SchemaObjectFactory` and its array-returning branch is reachable independently of the explorer: `test/services/schema-object-factory.spec.ts` exercises it directly in "should expand unnamed query standard schemas into parameter properties". Moving the expansion out would change behaviour for anyone calling `createFromModel` directly, so it would not be a pure refactor. Everything else from the review is addressed.

Issue #1818, opened 2022-02-11
Improve inference of @Param when used with ParseUUIDPipe


Delapouite:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Hi

Nest.js controller handlers often have to deal with UUIDs in the params. Here's an example:

```typescript
@Delete(`tenants/:tenantId/folders/:id`)
@ApiParam({ name: "tenantId", format: "uuid", type: "string" })
@ApiParam({ name: "id", format: "uuid", type: "string" })
async delete(
    @Param("tenantId", ParseUUIDPipe) tenantId: string,
    @Param("id", ParseUUIDPipe) id: string,
) {
    // …
}
```

As you can see, extra `@ApiParam` decorators are needed to help Swagger understand the fact that both `tenantId` and `id` params are expected to be uuid. But this information could be infer by the fact that we're using `ParseUUIDPipe` in `@Param`.

### Describe the solution you'd like

It would be great to not have to repeat the same information twice. While generating the Swagger document, the exploration process should be able to deduce that `@Param("id", ParseUUIDPipe)` means that the `id` param is an UUID.

Therefore, `@ApiParam` decorators would not be needed anymore, and the resulting code would be more lightweight, while providing the precise information in the final Swagger doc: 

```typescript
@Delete(`tenants/:tenantId/folders/:id`)
async delete(
    @Param("tenantId", ParseUUIDPipe) tenantId: string,
    @Param("id", ParseUUIDPipe) id: string,
) {
    // …
}
```

### What is the motivation / use case for changing the behavior?

DRY code.

Thanks

immint023:
```typescript

export function Id(name: string): any {
    return (target: any, key: string, descriptor: any) => {
        ApiParam({
            description: `${name} id`,
            example: 'Your custom example',
            name,
            required: false
        })(target, key, Object.getOwnPropertyDescriptor(target, key));
        return idParamDecorator(name)(target, key, descriptor);
    };
}

export const idParamDecorator = createParamDecorator((options: ParamOptions | string, ctx: ExecutionContext) => {
    let paramOptions = getParamOptions(options, 'id');
    const request = ctx.switchToHttp().getRequest();
    let id = request.params[paramOptions.key] || request.query[paramOptions.key];
    if (!id && paramOptions.nullable) {
        return id;
    }
    if (!isUUID(id)) {
        throw new UuidException(paramOptions.key);
    }
    return id;
[...]

PR #3864, opened 2026-04-21
feat(explorer): infer uuid format when ParseUUIDPipe is applied to a param

yogeshwaran-c:
## PR Checklist

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other

## Current behavior

When a controller handler uses `@Param('id', ParseUUIDPipe)` (or `@Query`, etc.), the generated OpenAPI schema only describes the parameter as `{ type: 'string' }`. In order to surface the UUID constraint in the docs, developers have to repeat the information with a matching `@ApiParam({ name: 'id', format: 'uuid', type: 'string' })` decorator:

```ts
@Delete('tenants/:tenantId/folders/:id')
@ApiParam({ name: 'tenantId', format: 'uuid', type: 'string' })
@ApiParam({ name: 'id', format: 'uuid', type: 'string' })
async delete(
  @Param('tenantId', ParseUUIDPipe) tenantId: string,
  @Param('id', ParseUUIDPipe) id: string,
) {}
```

## New behavior

The parameter metadata accessor now inspects the pipes attached to route argument decorators (which are already stored in `ROUTE_ARGS_METADATA` by Nest). When `ParseUUIDPipe` is detected - either as a class reference (`ParseUUIDPipe`) or as an instance (`new ParseUUIDPipe({ version: '4' })`) - the parameter is automatically emitted as `{ type: 'string', format: 'uuid' }` in the OpenAPI schema, so the redundant `@ApiParam` decorators are no longer required:

```ts
@Delete('tenants/:tenantId/folders/:id')
async delete(
  @Param('tenantId', ParseUUIDPipe) tenantId: string,
  @Param('id', ParseUUIDPipe) id: string,
) {}
```

An explicit `@ApiParam`/`@ApiQuery` decorator continues to take precedence, so existing documents are unaffected.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Fixes #1818

Issue #3191, opened 2024-11-26
ApiParam not overwriting type from @Param


dariozachow:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I'm trying to create a custom pipe which takes the id of an entity as an url param. The Pipe now checks if the entity exists returns it. The code of the pipe looks like this:

```typescript
@Injectable()
export class TransformAssetPipe implements PipeTransform<string, Promise<AssetEntity>> {
  constructor(private readonly assetService: AssetService) {}
  async transform(assetId: string, _metadata: ArgumentMetadata): Promise<AssetEntity> {
    const asset = await this.assetService.findOne(assetId);

    if (!asset) {
      throw new AssetNotFoundException(assetId);
    }

    return asset;
  }
}
```
Used by the following controller:

```typescript
 @Get('assets/:assetId')
  @ApiParam({type: "string", name: "assetId")
  async findOne(
    @Param('assetId', ParseUUIDPipe, TransformAssetPipe) asset: AssetEntity
  ): Promise<AssetEntity> {
    return asset
  }
```
Now the Swagger generates ULR-Params for each field of the entity 

![image](https://github.com/user-attachments/assets/ed9f14e7-dd24-41bd-8abb-524e497f662e)

### Minimum reproduction code

https://github.com/nestjs/swagger

### Expected behavior

@ApiParam should overwrite the type from the @Param so you can specify your own type if you want over the @ApiParam, but have a custom Mapping in a pipe.

### Package version

8.0.7

### NestJS version

10.4.1

### Node.js version

22

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

kamilmysliwiec:
Please provide a [minimum reproduction repository](https://minimum-reproduction.wtf/) (Git repository/[StackBlitz](https://stackblitz.com/fork/github/nestjs/typescript-starter)/[CodeSandbox](https://codesandbox.io/s/github/nestjs/typescript-starter/tree/master) project).

dariozachow:
https://stackblitz.com/edit/nestjs-typescript-starter-cjfte3?file=src%2Fapp.controller.ts and open die /swagger endpoint

kamilmysliwiec:
![image](https://github.com/user-attachments/assets/fbdc7a2d-77e5-4f09-905a-06982d05e85c)

dariozachow:
[...]

PR #4095, opened 2026-09-05
feat(parameters): infer the uuid format from ParseUUIDPipe

kavish-19:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

A route parameter bound with `ParseUUIDPipe` already declares that it is a UUID, but the generated document does not say so. The `format` has to be repeated in an explicit decorator:

```ts
@Delete('tenants/:tenantId/folders/:id')
@ApiParam({ name: 'tenantId', format: 'uuid', type: 'string' })
@ApiParam({ name: 'id', format: 'uuid', type: 'string' })
async delete(
  @Param('tenantId', ParseUUIDPipe) tenantId: string,
  @Param('id', ParseUUIDPipe) id: string
) {}
```

Issue Number: #1818

## What is the new behavior?

The `format` is inferred from the pipe, so the decorators are no longer needed to state it:

```ts
@Delete('tenants/:tenantId/folders/:id')
async delete(
  @Param('tenantId', ParseUUIDPipe) tenantId: string,
  @Param('id', ParseUUIDPipe) id: string
) {}
```

Both parameters now come out as `{ type: 'string', format: 'uuid' }`.

`ParameterMetadataAccessor` reads the `pipes` already recorded in `ROUTE_ARGS_METADATA` and emits `format: 'uuid'` when one of them is `ParseUUIDPipe`. Recognised forms are the pipe class, a pipe instance (`new ParseUUIDPipe({ version: '4' })`), and a subclass of it.

Precedence and scope:

- **An explicit `@ApiParam`/`@ApiQuery` format still wins**, because explicit parameters are merged over the reflected ones in `api-parameters.explorer.ts`.
- **Only string-typed parameters are considered.** A model-typed parameter is expanded into its properties downstream, so without that guard an inferred format would leak onto every property of the model.
- `format` needed no downstream changes — `SwaggerTypesMapper.getSchemaOptions()` already handles that key.

### Why this lives in the accessor

`createQueryOrParamSchema()` in `schema-object-factory.ts` is where format is currently inferred from a parameter's *type* (`Date` → `date-time`, `BigInt` → `int64`), and this change deliberately uses the same precedence ordering. It is not implemented there because the bound pipes are not carried on `ParamWithTypeMetadata`, and plumbing them through would push route-argument internals into the schema layer. Reading decorator metadata off the handler is what the accessor already does, so the inference sits next to the metadata it reads. Happy to move it if you would rather it lived with the type-driven inference.

- [ ] Yes
- [x] No

Not a breaking change to any API, but it does **change generated output for existing applications**: a project already using `ParseUUIDPipe` will see `format: 'uuid'` appear on those parameters where there was none before. That is the point of the feature, but it will show up as a diff for anyone generating clients from the document, so flagging it explicitly rather than letting it be discovered.

## Other information

Scoped to `ParseUUIDPipe` only, as #1818 asks. The same mechanism would extend to `ParseIntPipe`/`ParseBoolPipe`/`ParseFloatPipe` — happy to follow up if you want that, but I did not want to widen an issue-sized change unprompted.

Five tests added in `test/explorer/swagger-explorer.spec.ts`, covering the pipe class, a pipe instance, a query parameter alongside an untouched sibling parameter, explicit-format precedence, and a subclass. Each was confirmed to fail against the unmodified code before the fix.

Validated with:

```bash
npm test          # 46 files, 503 passed
npm run test:e2e  # 5 files, 133 passed
npm run build     # clean
npm run lint      # no new findings
npx prettier --check lib/services/parameter-metadata-accessor.ts test/explorer/swagger-explorer.spec.ts
```

https://claude.ai/code/session_0139tvDRWUjAP2JcJC4UeoPY

### DIFF
--- a/lib/services/parameter-metadata-accessor.ts
+++ b/lib/services/parameter-metadata-accessor.ts
@@ -1,10 +1,11 @@
-import { Type } from '@nestjs/common';
+import { PipeTransform, Type } from '@nestjs/common';
 import {
   PARAMTYPES_METADATA,
   ROUTE_ARGS_METADATA
 } from '@nestjs/common/constants.js';
 import { RouteParamtypes } from '@nestjs/common/enums/route-paramtypes.enum.js';
-import { isEmpty, mapValues, omitBy } from 'es-toolkit/compat';
+import { ParseUUIDPipe } from '@nestjs/common/pipes/parse-uuid.pipe.js';
+import { isEmpty, isFunction, mapValues, omitBy } from 'es-toolkit/compat';
 import { EnumSchemaAttributes } from '../interfaces/enum-schema-attributes.interface.js';
 import {
   ParameterLocation,
@@ -13,9 +14,12 @@ import {
 import { StandardSchemaObject } from '../interfaces/swagger-document-options.interface.js';
 import { reverseObjectKeys } from '../utils/reverse-object-keys.util.js';
 
+type ParamPipe = Type<PipeTransform> | PipeTransform;
+
 interface ParamMetadata {
   index: number;
   data?: string | number | object;
+  pipes?: ParamPipe[];
   schema?: StandardSchemaObject;
 }
 type ParamsMetadata = Record<string, ParamMetadata>;
@@ -27,6 +31,7 @@ export interface ParamWithTypeMetadata {
   standardSchema?: StandardSchemaObject;
   isArray?: boolean;
   items?: SchemaObject;
+  format?: string;
   required?: boolean;
   enum?: unknown[];
   enumName?: string;
@@ -36,6 +41,14 @@ export interface ParamWithTypeMetadata {
 export type ParamsWithType = Record<string, ParamWithTypeMetadata>;
 
 const PARAM_TOKEN_PLACEHOLDER = 'placeholder';
+
+function isParseUUIDPipe(pipe: ParamPipe): boolean {
+  return (
+    pipe === ParseUUIDPipe ||
+    pipe instanceof ParseUUIDPipe ||
+    (isFunction(pipe) && pipe.prototype instanceof ParseUUIDPipe)
+  );
+}
 
 export class ParameterMetadataAccessor {
   explore(
@@ -65,7 +78,8 @@ export class ParameterMetadataAccessor {
           type: types[param.index],
           name: param.data,
           standardSchema: param.schema,
-          required: true
+          required: true,
+          ...this.inferSchemaFromPipes(types[param.index], param.pipes)
         }) as unknown as ParamWithTypeMetadata
     ) as unknown as ParamsWithType;
     const excludePredicate = (val: ParamWithTypeMetadata) =>
@@ -79,6 +93,22 @@ export class ParameterMetadataAccessor {
       excludePredicate as Function
     );
     return !isEmpty(parameters) ? (parameters as ParamsWithType) : undefined;
+  }
+
+  /**
+   * Infers schema attributes that a bound pipe already guarantees, so that
+   * they don't have to be repeated in an explicit `@ApiParam`/`@ApiQuery`.
+   * Only string-typed parameters are considered, to avoid leaking the
+   * inferred attributes onto the properties of a model-typed parameter.
+   */
+  private inferSchemaFromPipes(
+    type: Type<unknown>,
+    pipes: ParamPipe[] = []
+  ): Pick<ParamWithTypeMetadata, 'format'> | undefined {
+    if (type !== String || !pipes.some(isParseUUIDPipe)) {
+      return undefined;
+    }
+    return { format: 'uuid' };
   }
 
   private mapParamType(key: string): string {
--- a/test/explorer/swagger-explorer.spec.ts
+++ b/test/explorer/swagger-explorer.spec.ts
@@ -4,6 +4,7 @@ import {
   Controller,
   Get,
   Param,
+  ParseUUIDPipe,
   Post,
   Query,
   Version,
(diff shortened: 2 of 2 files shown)


lib/services/parameter-metadata-accessor.ts
```diff
@@ -1,10 +1,11 @@
-import { Type } from '@nestjs/common';
+import { PipeTransform, Type } from '@nestjs/common';
 import {
   PARAMTYPES_METADATA,
   ROUTE_ARGS_METADATA
 } from '@nestjs/common/constants.js';
 import { RouteParamtypes } from '@nestjs/common/enums/route-paramtypes.enum.js';
-import { isEmpty, mapValues, omitBy } from 'es-toolkit/compat';
+import { ParseUUIDPipe } from '@nestjs/common/pipes/parse-uuid.pipe.js';
```
micalevisk, line 7: Please import this from the public barrel, alongside the `PipeTransform, Type` import on line 1:

```ts
import { ParseUUIDPipe, PipeTransform, Type } from '@nestjs/common';
```

In `lib/` the deep `@nestjs/common/...` paths are only used for internals that are not re-exported (`constants.js`, `utils/shared.utils.js`, the `RouteParamtypes` enum). `ParseUUIDPipe` is `@publicApi`, and the test file already imports it from `@nestjs/common`. Deep paths also tie us to the package's internal layout.

lib/services/parameter-metadata-accessor.ts
```diff
@@ -37,6 +42,14 @@ export type ParamsWithType = Record<string, ParamWithTypeMetadata>;
 
 const PARAM_TOKEN_PLACEHOLDER = 'placeholder';
 
+function isParseUUIDPipe(pipe: ParamPipe): boolean {
+  return (
+    pipe === ParseUUIDPipe ||
+    pipe instanceof ParseUUIDPipe ||
+    (isFunction(pipe) && pipe.prototype instanceof ParseUUIDPipe)
+  );
+}
```
micalevisk, line 51: nit: the first two checks fold into the third once you branch on "class vs instance":

```ts
function isParseUUIDPipe(pipe: ParamPipe): boolean {
  return isFunction(pipe)
    ? pipe === ParseUUIDPipe || pipe.prototype instanceof ParseUUIDPipe
    : pipe instanceof ParseUUIDPipe;
}
```

Not blocking, current version is correct.

lib/services/parameter-metadata-accessor.ts
```diff
@@ -81,6 +95,22 @@ export class ParameterMetadataAccessor {
...
+   * Infers schema attributes that a bound pipe already guarantees, so that
+   * they don't have to be repeated in an explicit `@ApiParam`/`@ApiQuery`.
+   * Only string-typed parameters are considered, to avoid leaking the
+   * inferred attributes onto the properties of a model-typed parameter.
+   */
+  private inferSchemaFromPipes(
+    type: Type<unknown>,
+    pipes: ParamPipe[] = []
+  ): Pick<ParamWithTypeMetadata, 'format'> | undefined {
+    if (type !== String || !pipes.some(isParseUUIDPipe)) {
```
micalevisk, line 108: This guard is what makes the difference vs. #3829/#3864: with `@Param('id', ParseUUIDPipe, EntityLookupPipe) entity: AssetEntity` (the multi-pipe scenario Kamil raised, and a real one from #3191) the reflected type is the entity class, so nothing is inferred. Worth calling out in the PR description if you want to reopen the discussion.

One case it does not cover: an explicit `@ApiParam({ name: 'id', type: Number })` on a `ParseUUIDPipe` param comes out as `{ type: 'number', format: 'uuid' }`, because the explicit merge is per-key. Not asking to fix it here, just noting it.

test/explorer/swagger-explorer.spec.ts
```diff
@@ -2608,6 +2609,139 @@ describe('SwaggerExplorer', () => {
     });
   });
 
+  describe('when ParseUUIDPipe is used', () => {
```
micalevisk, line 2612: The five cases here are good, but the `type !== String` guard, which the description calls out as the reason the inferred format can't leak onto model properties, has no test. Please add a case like:

```ts
class FilterDto {
  @ApiProperty() a: string;
  @ApiProperty() b: number;
}

@Get('graults')
findGrault(@Query(ParseUUIDPipe) query: FilterDto) {}
```

asserting that both `a` and `b` come out without `format`. (I ran it locally and it passes, so this is purely about locking the behaviour in.)

Since the logic lives in the accessor, a focused case in `test/services/parameter-metadata-accessor.spec.ts` asserting `format: 'uuid'` on the reflected metadata would also be a natural home, with the explorer tests kept as the end-to-end check. Optional.

micalevisk (changes requested):
Thanks for the PR, and for the thorough description. Before anything else I have to surface some prior history, because it decides how this PR can move forward.

### This feature has been declined twice before

The same change (reading `pipes` from `ROUTE_ARGS_METADATA` in `ParameterMetadataAccessor` and emitting `format: 'uuid'` for `ParseUUIDPipe`) was submitted and closed by @kamilmysliwiec earlier this year:

- #3829 (April 2026), closed with: *"Feels a bit hacky. What if someone has multiple pipes applied? I'd say users should just specify this manually"*
- #3864 (April 2026), closed with: *"feels a bit implicit and hacky isn't it?"*

This PR is the third attempt at the same design (it even shares the `inferSchemaFromPipes` name and the `pipes?`/`format?` interface additions with #3864). It is a cleaner implementation than both, but the objection was to the idea of inferring OpenAPI schema from runtime pipes, not to the implementation details, so I can't approve it on my own. The precedent in this repo is that inference comes from TypeScript types (`Date` → `date-time`, `BigInt` → `int64`, node's `UUID` template type in #2797), not from pipes.

If you want to make the case, the review would need a direct answer to the "multiple pipes" objection. For example: `@Param('id', ParseUUIDPipe, EntityLookupPipe) entity: AssetEntity` (a real case from #3191). Your `type !== String` guard happens to skip that one because the reflected type is the entity class, which is a partial mitigation nobody had before, and that is worth stating explicitly. Also #1818 itself has never had a maintainer response, so the issue being open is not a signal that the feature is wanted. I'm leaving the final call to Kamil.

### Implementation review (in case the design is accepted)

I checked out the branch: unit (503) / e2e (133) / lint / build / prettier are green locally, and I traced `format` end-to-end (`ParameterMetadataAccessor` → `ParametersMetadataMapper.transformModelToProperties` → `SchemaObjectFactory.createFromModel` → `SwaggerTypesMapper.mapParamTypes`). The precedence claims hold:

- explicit `@ApiParam({ format })` overrides the inferred one (per-key `assign` in `api-parameters.explorer.ts`) ✅
- explicit `@ApiParam({ schema })` drops it entirely (`format` is in `keysToRemove`) ✅
- the `type !== String` guard prevents the leak onto model properties (`@Query(ParseUUIDPipe) q: SomeDto` still expands to plain properties) ✅

Two small requests, see inline comments:

1. Import `ParseUUIDPipe` from the `@nestjs/common` barrel instead of the deep `pipes/parse-uuid.pipe.js` path.
2. Add a test for the `type !== String` guard. It's the load-bearing safety check of this change and it's currently untested.

### Gaps I found while probing (follow-ups, not blocking)

- **Explicit `type` doesn't clear the inferred `format`.** `@ApiParam({ name: 'id', type: Number })` + `@Param('id', ParseUUIDPipe)` currently yields `schema: { type: 'number', format: 'uuid' }`. "Explicit wins" is per-key today, so the inferred format survives a contradicting explicit `type`.
- **`new ParseUUIDPipe({ optional: true })`** on a `@Query()` still emits `required: true`. Pre-existing behaviour for all reflected params, but if we read the pipe we could infer `required: false` from it too.
- **Docs:** if this lands, `content/openapi/types-and-parameters.md` in https://github.com/nestjs/docs.nestjs.com should mention it.

### On the output change

Agree it would not be a breaking change: `format: uuid` is additive and only appears where the runtime already rejects non-UUIDs.

kamilmysliwiec:
what if someone has another piped inlined that converts uuid to something else later?

PR #4110, opened 2026-09-12
fix: avoid duplicate parameters when combining standard schema with @ApiQuery

Asadshah7950:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When combining a Standard Schema parameter (e.g. `@Query({ schema: z.object({ limit: ... }) })`) with an explicit parameter decorator such as `@ApiQuery({ name: 'limit', description: 'Max results', required: false })`, the generated OpenAPI document emits **two duplicate parameter entries** for `limit` in `query`:
```json
[
  {
    "name": "limit", "in": "query", "required": false,
    "schema": { "type": "integer", "minimum": 1, "maximum": 100 }
  },
  {
    "name": "limit", "in": "query", "required": false,
    "description": "Max results",
    "schema": {}
  }
]
```

This occurs because:
1. Reflected parameters with `standardSchema` initially have `name: undefined` when `unionWith` / `mergeImplicitAndExplicit` runs in `exploreApiParametersMetadata`.
2. Property expansion occurs later during `schemaObjectFactory.createFromModel(...)`.
3. Consequently, the identity check (`hasSameParameterIdentity`) misses the correlation prior to expansion, resulting in duplicate parameter definitions sharing the same `(name, in)` identity, which is invalid per the OpenAPI 3.0/3.1 specification.

Issue Number: #4088

## What is the new behavior?
`exploreApiParametersMetadata` now deduplicates and merges parameters sharing the same `(name, in)` identity after model creation:
- Preserves the detailed schema definition generated from Standard Schema validation (types, constraints, min/max).
- Merges explicit metadata from decorators (`description`, `example`, `examples`, `deprecated`, etc.).
- Guarantees valid OpenAPI operations with unique parameters by `(name, in)`.

Unit tests covering `@ApiQuery` combined with Standard Schema query parameters have been added to `test/explorer/swagger-explorer.spec.ts`.

- [ ] Yes
- [x] No

## Other information
All 47 test files and 500 tests pass cleanly.


micalevisk (changes requested):
Thanks for picking this up. I ran the branch locally (unit + e2e green), but I'd like a few changes before merging. The main concern is that the fix is applied at the wrong layer and the schema-merge rule is non-deterministic.

### 1. The schema merge rule ("whichever side has more keys wins") produces invalid schemas

`deduplicateParameters` picks the base schema by comparing `Object.keys(...).length`. That makes the result depend on how many constraints each side happens to carry. I probed three small variants of the fixture against this branch:

| `@ApiQuery(...)` | Zod property | Resulting `schema` |
|---|---|---|
| `{ type: String, enum: ['a','b'] }` | `z.number().int().min(1)` | `{ enum: ['a','b'], type: 'integer', minimum: 1, … }` — string enum on an integer |
| `{ type: Number, minimum: 0, maximum: 10, default: 5 }` | `z.string().min(2)` | `{ type: 'number', minLength: 2, minimum: 0, … }` — explicit side won (4 keys vs 2) and the runtime validator's type was discarded |
| `{ type: Number }` | `z.string().min(2).max(5)` | `{ type: 'string', minLength: 2, maxLength: 5 }` — Zod side won (3 keys vs 1), correct |

The PR description says the Standard Schema definition is preserved, but the code doesn't guarantee that. We need one deterministic rule. The expectation stated in #4088 is *"the same merge class-DTO query parameters already get"*, i.e. explicit metadata is assigned on top of the reflected param (`assign(item, explicit)`) while the reflected schema stays the source of truth. The existing `standard-query-conflict` e2e assertion already encodes that expectation (the `filter` param must keep the Zod integer schema).

### 2. Please fix the ordering problem instead of deduplicating after the fact

Class DTOs are expanded into named properties in `ParametersMetadataMapper.transformModelToProperties` **before** `hasSameParameterIdentity` runs, which is why they merge correctly. Standard Schema params short-circuit there (`if (param.standardSchema) return param;`) and are only expanded later in `SchemaObjectFactory.createSchemaOverrideParam`. That ordering is the root cause.

A global dedupe after `mapParamTypes` papers over it and introduces a second merge path with its own semantics next to the existing `mergeImplicitAndExplicit`. It also reimplements the identity check inline instead of reusing `hasSameParameterIdentity` (and drops its body/body special case). Expanding Standard Schema params before the identity check, or running the existing merge against the expanded output, fixes it in one place and keeps a single set of merge semantics.

### 3. Regenerate `e2e/api-spec.json`

#4088 points at the `standard-query-conflict` route in `e2e/src/cats/cats.controller.ts`; the duplicate `filter` entry is currently committed in `e2e/api-spec.json`. Running `npm run test:e2e` on this branch rewrites that block (duplicate removed, `required` becomes `false`, description merged). That diff needs to be part of the PR, otherwise the fixture contradicts the code.

(You'll also see two unrelated hunks move — an email `pattern` and an enum key order. Those drift identically on `master`, so please leave them out of this PR.)

### 4. Add an e2e uniqueness assertion

The e2e helpers are `.find()`-based, so they cannot detect duplicates — which is exactly why this shipped unnoticed. Please add an assertion on the conflict route that parameters are unique by `(name, in)`. The new unit test only covers the happy path from the issue; at least one of the conflicting-schema cases from the table above should be covered too, whichever rule we settle on.

### Nits

- `npm run format` — the `return` on line 97 of `api-parameters.explorer.ts` exceeds the print width.
- The new function takes/returns `any[]`; `ReturnType<SwaggerTypesMapper['mapParamTypes']>` is available.
- If a post-processing pass does stay, it runs an O(n²) `findIndex` on every operation, including the vast majority without any Standard Schema. Guard it on `some(properties, p => !!p.standardSchema)` (duplicates can only originate from Standard Schema expansion since `unionWith` already dedupes everything else), or key a `Map` on `${in}:${name}`.
- `{ ...existing, ...param }` appends `description` after `schema`, whereas every other parameter in the document has `description` before `schema` (visible in the regenerated fixture).
- `import { z } from 'zod'` in `swagger-explorer.spec.ts` should go with the other third-party imports, above the relative ones.

Asadshah7950:
Thank you for the detailed and insightful review @micalevisk!

All requested changes have been implemented:

1. **Ordering & Single Merge Pipeline**:
   - Standard Schema non-body query/param properties are now expanded into named parameter definitions via `schemaObjectFactory.createFromModel` **before** `hasSameParameterIdentity` runs, aligning them directly with the class-DTO parameter expansion flow.
   - Removed `deduplicateParameters` entirely and reused the existing `hasSameParameterIdentity` / `mergeImplicitAndExplicit` / `unionWith` pipeline.
2. **Deterministic Merge Rule**:
   - Explicit decorator metadata (`description`, `required`, etc.) is assigned onto the parameter while strictly preserving the reflected schema as the source of truth (`merged.schema = reflectedSchema`).
   - Ensured parameter-level properties like `description` are placed before `schema`.
3. **Regenerated `e2e/api-spec.json`**:
   - Regenerated via `npm run test:e2e` for `/api/cats/standard-query-conflict` (duplicate removed, `required: false`, description merged). Left out the unrelated email pattern and enum order drift.
4. **Uniqueness Assertion & Tests**:
   - Added `expect(parameterKeys).toEqual(Array.from(new Set(parameterKeys)))` in `e2e/standard-schema.e2e-spec.ts` to assert parameter uniqueness by `(in, name)`.
   - Added a unit test in `test/explorer/swagger-explorer.spec.ts` verifying that conflicting `@ApiQuery` type/enum definitions do not overwrite the reflected standard schema.
5. **Nits & Imports**:
   - Moved `import { z } from 'zod'` to third-party imports and verified all linting and formatting. All 500 unit tests and 133 e2e tests pass cleanly.

micalevisk (approved):

kamilmysliwiec:
Superseded by #4090, which fixes #4088 by expanding Standard Schema params before the identity merge rather than deduping after `mapParamTypes` — the approach requested in review here. That is now merged.

Thanks for digging into this. One piece of this PR is still worth keeping on its own: the `omitBy` → `.filter` change in `removeBodyMetadataIfExplicitExists`. `omitBy` from `es-toolkit/compat` returns a plain object when handed an array, so the `as ParamWithTypeMetadata[]` cast there is wrong and everything downstream operates on an object instead of an array. That is an existing bug on `master`, independent of the duplication fix — very happy to take it as its own PR if you would like to open one.

Issue #4116, opened 2026-09-14
CLI plugin emits an extensionless node_modules import under esmCompatible, producing a build that cannot be loaded (ERR_MODULE_NOT_FOUND)


thetimmytook:
### Is there an existing issue for this?

Searched; #4079 (fixed by #4082 in 12.0.1) is the other half of the same branch (`await import` emitted without `async`), not this.

### Current behavior

With `"type": "module"` in `package.json` and the CLI plugin enabled, the plugin hoists a static namespace import for every out-of-file type it needs in the generated metadata factory. For a type whose declaration lives inside a dependency, the emitted specifier keeps the declaration-file path **without an extension**:

```js
// dist/foo/foo.dto.js — emitted by the plugin, present in no source file
import * as openapi_import_1 from "legacy-pkg/dist/cabin-class.enum";
```

CommonJS `require` accepts that specifier; Node's ESM resolver does not. So `nest build` is green, `tsc` is green, eslint is green — and the `dist` it produces throws on boot:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../node_modules/legacy-pkg/dist/cabin-class.enum'
imported from .../dist/foo/foo.dto.js
Did you mean to import "legacy-pkg/dist/cabin-class.enum.js"?
```

Nothing in the toolchain catches it: the plugin is a `before` transformer, so it runs *after* type-checking and the emitted import is never type-checked; the test runner compiles the sources itself and never runs the plugin; lint reads `src`. The first signal is the process exiting at runtime.

### Expected behavior

The emitted specifier resolves under Node ESM — i.e. the plugin appends the extension of the resolved file for `node_modules` specifiers too, the same way it already does for project-relative ones.

### Minimal reproduction

1. A dependency with **no `exports` map** whose root `.d.ts` re-exports from a deep file, e.g. `legacy-pkg` with `dist/cabin-class.enum.d.ts` declaring `export enum CabinClass { Economy = 'Economy' }` and `dist/index.d.ts` re-exporting it. (Any `.d.ts` layout where TypeScript's `import("…")` type reference names the deep path reproduces it.)
2. A Nest app with `"type": "module"`, `tsconfig` `module`/`moduleResolution` `nodenext`, `nest-cli.json` → `"plugins": ["@nestjs/swagger/plugin"]`.
3. A DTO property typed by that enum:

```ts
import { CabinClass } from 'legacy-pkg';

export class FooDto {
[...]

PR #4120, opened 2026-09-15
fix(plugin): give a node_modules subpath its extension under esm

johnnyhuirilef:
## PR Checklist

- [x] The commit message follows our guidelines

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #4116

The plugin folds a declaration path back into a package subpath. A package
without an `exports` map exposes that subpath as it sits on disk, and Node's
ESM resolver needs the extension it carries.

```js
import * as openapi_import_1 from "legacy-pkg/dist/cabin-class.enum";
// ERR_MODULE_NOT_FOUND
```

The branch that appends the extension is only reached when
`normalizePackagePath` did *not* rewrite the path, which is never the case for
`node_modules`. The build stays green and the failure shows up at boot.

## What is the new behavior?

The extension comes from the file the dependency declares the type in, which
TypeScript has already resolved:

```
types.d.ts   -> .js
types.d.mts  -> .mjs
types.d.cts  -> .cjs
```

The specifier itself does not change: the subpath `normalizePackagePath`
produces is still the right one, as already settled earlier. Only the
extension is added, and only under `esmCompatible`.

Picking the specifier and picking the extension are now two steps rather than
one nested inside the other, so the chain stays about the specifier and the
extension rule reads in one place:

```ts
const isPackageSubpath = normalizedPath !== relativePath;
if (isPackageSubpath) {
  relativePath = normalizedPath;
} else if (sourceSpecifier) {
  relativePath = sourceSpecifier;
}

if (options.esmCompatible && relativePath !== sourceSpecifier) {
  relativePath += isPackageSubpath
    ? getDependencyExtension(declarationFileName)
    : getOutputExtension(fileName);
}
```

- [ ] Yes
- [x] No

## Other information

**On the shape.** `replaceImportPath` takes a fifth parameter,
`declarationFileName`, following how `sourceSpecifier` was added. The
alternative was passing the `ts.Type` and resolving inside, but that would pull
the TypeScript AST into a module that only handles strings today.

That said: five positional parameters is already a lot, and an options object
would read better. I am not proposing it here because it changes the signature
for existing callers, which is a separate decision from this fix.

**Why it does not probe the disk.** The module does no I/O today, and the tests
use paths that do not exist. TypeScript has already resolved the file, so
reading its name is enough.

**Tests.** Three in `plugin-utils.spec.ts`: the subpath gets `.js`, `.d.mts`
and `.d.cts` keep their own extension, and CommonJS output is untouched. The
first fails without the change.


lib/plugin/utils/plugin-utils.ts
```diff
@@ -347,10 +364,15 @@ export function replaceImportPath(
...
-      // Add appropriate extension for non-node_modules imports
-      const extension = getOutputExtension(fileName);
-      relativePath += extension;
+    }
+
+    if (options.esmCompatible && relativePath !== sourceSpecifier) {
+      // A package subpath names a file inside the dependency, so it carries
+      // that file's extension; a project-relative path carries the emitting
+      // file's. A source specifier already resolves and is left alone.
+      relativePath += isPackageSubpath
```
kamilmysliwiec, line 373: This runs for every normalized `node_modules` path, including ones `normalizePackagePath` collapsed at `/index`:

- `node_modules/pkg/index` → `pkg` → `pkg.js` (was `pkg`, which resolved via `main`/`exports`)
- `node_modules/pkg/dist/index` → `pkg/dist.js`

It also applies to packages that *do* have an `exports` map (where the extensionless subpath may be the one that's exported), to `@types/*` paths, and to paths that already end in `.js`/`.mjs`/`.cjs`. Could we only add the extension when the extensionless specifier doesn't resolve?

lib/plugin/utils/plugin-utils.ts
```diff
@@ -254,11 +254,27 @@ export function getOutputExtension(fileName: string): string {
   }
 }
 
+/**
+ * Gives the runtime extension of the file a dependency declares a type in.
+ * A package without an "exports" map exposes the subpath as it sits on disk,
+ * and Node's ESM resolver needs the extension that subpath carries.
+ */
+function getDependencyExtension(declarationFileName?: string): string {
```
kamilmysliwiec, line 262: nit: `getOutputExtension` already does this – `'x.d.mts'.endsWith('.mts')` / `'x.d.cts'.endsWith('.cts')` are both true, so `getOutputExtension(declarationFileName ?? '')` returns the same values and this helper can go.

lib/plugin/utils/type-reference-to-identifier.util.ts
```diff
@@ -141,6 +142,15 @@ function resolveSourceSpecifier(
   return symbol ? sourceImportSpecifiers.get(symbol) : undefined;
 }
 
+/**
+ * Returns the file a dependency declares `type` in. Its extension tells which
+ * file the package ships, which the type reference alone does not carry.
+ */
+function resolveDeclarationFileName(type: ts.Type): string | undefined {
+  const symbol = type?.aliasSymbol ?? type?.symbol;
```
kamilmysliwiec, line 150: For array properties `type` is the `Array<T>` type here, so this resolves to `lib.es5.d.ts` and always yields `.js` – e.g. `status: CabinClass[]` with `CabinClass` in a `.d.mts` gets `.js` instead of `.mjs`. `resolveSourceSpecifier` above unwraps by `arrayDepth`; the same should happen here (maybe share the unwrap).

Also, `declarations[0]` may be in a different file than the one in the type reference for merged declarations/augmentations – picking the declaration whose source file matches the import path would be safer.

kamilmysliwiec (commented):
Thanks for tackling this! The deep-subpath case is fixed, but appending an extension to every `node_modules` subpath regresses a few common cases (details inline):

1. **Package root / `index` imports** – `normalizePackagePath` truncates at `/index`, so a type from `node_modules/pkg/index.d.ts` now emits `pkg.js` instead of `pkg` (and `pkg/dist/index` → `pkg/dist.js`). This is probably the most common case and would fail at boot with `ERR_MODULE_NOT_FOUND`.
2. **Packages with an `exports` map** – the description scopes the fix to packages *without* `exports`, but the extension is added unconditionally. A package mapping e.g. `"./dist/*": "./dist/*.js"` resolved `pkg/dist/types` before and now gets `pkg/dist/types.js` → `./dist/types.js.js`.
3. **Array types** – `resolveDeclarationFileName` doesn't unwrap arrays (unlike `resolveSourceSpecifier`), so for `Enum[]` it reads `Array` from `lib.es5.d.ts` and always yields `.js`, even when the enum lives in a `.d.mts`.

Smaller things: a subpath that already carries an extension would get a second one; `@types/foo/...` gets `.js` assuming the runtime package mirrors the typings layout; `declarations[0]` may point to a different file for merged declarations.

Since all of these stem from building the specifier via string manipulation, it might be more robust to only append the extension when the extensionless specifier doesn't resolve (e.g. try resolving the candidate from the emitting file, and keep it as-is when it does).

Tests: it'd be great to add cases for `import("/repo/node_modules/pkg/index").Foo` with `esmCompatible: true` (expect `pkg`) and an array-of-`.d.mts`-enum property through `typeReferenceToIdentifier`. Also, the unrelated prettier reformatting in `plugin-utils.spec.ts` would be better left out of this PR to keep the diff focused.

johnnyhuirilef:
Reworked 🙏 You were right on all four, and I could reproduce each one:

```
node_modules/pkg/index       -> pkg      -> pkg.js      (would not resolve)
node_modules/pkg/dist/index  -> pkg/dist -> pkg/dist.js (would not resolve)
```

The extension is now appended only where the specifier does not already reach a file, as you suggested. Two things I measured along the way, in case they are useful:

Resolution alone is not enough. `import.meta.resolve` hands back a URL without looking at what is there, so an extensionless subpath of a package without an `exports` map still "resolves" — to a path that is not on disk. A directory answers too, which is what `/index` being dropped upstream produces. Both need the `statSync().isFile()`.

`getDependencyExtension` is gone, as you noted. And `resolveDeclarationFileName` now unwraps arrays through the same helper as `resolveSourceSpecifier`, so an enum in a `.d.mts` behind `Enum[]` gets `.mjs`.

The tests run against a package fixture rather than installed dependencies, one package per branch: no `exports` map, an `exports` map, one publishing both forms, a directory, and a declaration with no runtime file beside it. That needs one `.gitignore` negation, since the ESM resolver only reads packages from a folder named `node_modules` — happy to take a different route there if you prefer.

I also dropped the unrelated prettier reformatting from the spec.

johnnyhuirilef:
Correction to my last comment: when I said the prettier reformatting was out, the diff still carried it — a `--write` over the whole file had reflowed lines that were not mine, and a rename inside my fixture had leaked into two unrelated tests.

Both are gone now. The spec is `+102/-0`, and the only lines removed anywhere are the ones this change replaces.

kamilmysliwiec:
Thanks for the rework! I pushed two follow-up commits before merging:

- `import.meta.resolve` ignores its `parent` argument unless Node runs with `--experimental-import-meta-resolve`, so outside vitest resolution started from the plugin's own location and the extension was never appended (`plain-pkg/out/status` stayed extensionless in plain Node). The check now reads the package from disk: for a package without `exports`, the extension is appended only when the literal subpath is not a file and the path with the extension is. Packages with an `exports` map and package roots are left as they are.
- The declaration file is now picked by matching the `import("...")` path, instead of `declarations[0]`, so merged declarations and augmentations get the right extension.
- Added an end-to-end spec (`esm-package-subpath.spec.ts`) covering a package root, a `.d.mts` enum, a `.d.mts` class array, and a check that the emitted specifiers resolve in plain Node.

PR #4125, opened 2026-09-16
feat(swagger): infer parameter type and format from pipes

Asadshah7950:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

When parameters in controller handlers are configured with standard NestJS transformation/validation pipes (e.g. `@Param('id', ParseUUIDPipe)` or `@Param('count', ParseIntPipe)`), developers still have to explicitly add `@ApiParam({ format: 'uuid' })` or `@ApiParam({ type: Number })` for OpenAPI parameter schemas to reflect the format or type.

Issue Number: #1818

## What is the new behavior?

`ParameterMetadataAccessor` now inspects the `pipes` attached to route parameters in `ROUTE_ARGS_METADATA`:
- `ParseUUIDPipe` -> infers `type: String` and `format: 'uuid'`
- `ParseIntPipe` -> infers `type: Number`
- `ParseFloatPipe` -> infers `type: Number`
- `ParseBoolPipe` -> infers `type: Boolean`

Explicit `@ApiParam()` and `@ApiQuery()` definitions continue to take precedence.

- [ ] Yes
- [x] No


lib/services/parameter-metadata-accessor.ts
```diff
@@ -17,12 +17,14 @@ interface ParamMetadata {
   index: number;
   data?: string | number | object;
   schema?: StandardSchemaObject;
+  pipes?: any[];
```
micalevisk, line 20: nit: this can be typed instead of `any[]`:

```ts
  pipes?: (Type<PipeTransform> | PipeTransform)[];
```
(with `PipeTransform` imported from `@nestjs/common`)

lib/services/parameter-metadata-accessor.ts
```diff
@@ -60,13 +62,19 @@ export class ParameterMetadataAccessor {
 
     const parametersWithType: ParamsWithType = mapValues(
       reverseObjectKeys(routeArgsMetadata),
-      (param: ParamMetadata) =>
-        ({
-          type: types[param.index],
+      (param: ParamMetadata) => {
+        const pipeInferences = this.inferFromPipes(param.pipes);
+        const paramMetadata: ParamWithTypeMetadata = {
+          type: pipeInferences.type ?? types[param.index],
```
micalevisk, line 68: Here the pipe wins over the reflected TypeScript type. That is what changes the generated document for existing apps (e.g. `@Query("q", ParseFloatPipe) q: string` goes from `string` to `number`, and `id: string | number` params that were omitted before now appear). #4095 deliberately only *added* `format` when the reflected type was already `String`, which was the more conservative choice; this goes in the opposite direction of the feedback given there.

lib/services/parameter-metadata-accessor.ts
```diff
@@ -96,4 +104,27 @@ export class ParameterMetadataAccessor {
         return PARAM_TOKEN_PLACEHOLDER;
     }
   }
+
+  private inferFromPipes(pipes?: any[]): { type?: Type<unknown>; format?: string } {
```
micalevisk, line 108: `prettier --check` fails on this file because of this line (> 80 cols). Please run `npm run format`.

lib/services/parameter-metadata-accessor.ts
```diff
@@ -96,4 +104,27 @@ export class ParameterMetadataAccessor {
...
+    if (!Array.isArray(pipes)) {
+      return {};
+    }
+    for (const pipe of pipes) {
```
micalevisk, line 112: This is exactly the case raised in #3829 and #4095: the loop returns on the first *matching* pipe regardless of its position in the chain. Only the first pipe sees the wire value, so for `@Param("slug", SlugToIdPipe, ParseIntPipe)` the documented input would be `number` while the API actually accepts a slug string. And for `@Param("id", ParseUUIDPipe, EntityLookupPipe) entity: AssetEntity` (real case from #3191) the reflected type is discarded. At the very least only `pipes[0]` (skipping `DefaultValuePipe`) could be trusted to describe the input, but that is part of why this was considered too implicit.

lib/services/parameter-metadata-accessor.ts
```diff
@@ -96,4 +104,27 @@ export class ParameterMetadataAccessor {
...
+      const ctor = typeof pipe === 'function' ? pipe : pipe?.constructor;
+      const name = ctor?.name;
```
micalevisk, line 114: Matching on `constructor.name` is fragile. Verified locally:

- a user-land `class ParseIntPipe implements PipeTransform` (different semantics, same name) is now documented as `number` (false positive)
- `class MyIntPipe extends ParseIntPipe {}` is not detected (false negative)
- minified/mangled builds silently lose the inference

`@nestjs/common` is already a peer dependency imported by this file, so the real classes can be used: `pipe === ParseIntPipe || pipe instanceof ParseIntPipe || pipe?.prototype instanceof ParseIntPipe`.

lib/services/parameter-metadata-accessor.ts
```diff
@@ -96,4 +104,27 @@ export class ParameterMetadataAccessor {
...
+      if (name === 'ParseUUIDPipe') {
+        return { type: String as unknown as Type<unknown>, format: 'uuid' };
```
micalevisk, line 116: Two things:

1. Because `format` is set independently from `type`, `@ApiParam({ name: "id", type: Number })` + `@Param("id", ParseUUIDPipe)` now produces `{ "type": "number", "format": "uuid" }` (verified locally). Same problem I pointed out in #4095.
2. The `as unknown as Type<unknown>` casts are not needed, `String`/`Number`/`Boolean` are assignable to `Type<unknown>` (checked with `tsc`). The `if` chain could also be a small lookup table (`[[ParseUUIDPipe, { type: String, format: "uuid" }], ...]`) so adding a pipe is one line.

lib/services/parameter-metadata-accessor.ts
```diff
@@ -96,4 +104,27 @@ export class ParameterMetadataAccessor {
...
+      }
+      if (name === 'ParseIntPipe') {
```
micalevisk, line 118: `ParseIntPipe` rejects `1.5`, so the accurate OpenAPI type is `integer`, not `number`. Emitting `number` here makes it identical to `ParseFloatPipe`, which loses the only information the pipe adds over a `number` TS annotation.

test/services/parameter-metadata-accessor.spec.ts
```diff
@@ -42,4 +50,46 @@ describe('ParameterMetadataAccessor', () => {
       ])
     );
   });
+
+  it('should infer type and format from NestJS pipes', () => {
```
micalevisk, line 54: Coverage gaps: `ParseFloatPipe` (implemented but never tested), pipe instances (`new ParseIntPipe()`, `new ParseUUIDPipe({ version: "4" })`), several pipes on the same param, `@Query()`, and a param whose TS type disagrees with the pipe (since the pipe overrides it). Also there is a stray blank line at the end of this file that fails `prettier --check`.

test/explorer/swagger-explorer.spec.ts
```diff
@@ -3244,4 +3247,60 @@ describe('SwaggerExplorer', () => {
...
+      get(
+        @Param('tenantId', ParseUUIDPipe) tenantId: string,
+        @Param('id', ParseIntPipe) id: number,
+        @Query('active', ParseBoolPipe) active: boolean
+      ) {
+        return true;
+      }
+    }
+
+    it('infers type and format from pipes without explicit @ApiParam/@ApiQuery', () => {
```
micalevisk, line 3264: The PR description states that explicit `@ApiParam()`/`@ApiQuery()` keep precedence, but nothing asserts it. Please add cases for an explicit `type`, `enum`, `schema` and a description-only `@ApiParam()` on piped params. The `type: Number` + `ParseUUIDPipe` combination currently yields `{ type: "number", format: "uuid" }`.

micalevisk (changes requested):
Thanks for the PR! Before the code itself, I have to surface the history of this feature, because it decides how this can move forward.

### This feature has already been declined three times

The same design (reading `pipes` from `ROUTE_ARGS_METADATA` in `ParameterMetadataAccessor` and deriving schema from them) was submitted and closed by @kamilmysliwiec:

- #3829 (April 2026): *"Feels a bit hacky. What if someone has multiple pipes applied? I'd say users should just specify this manually"*
- #3864 (April 2026): *"feels a bit implicit and hacky isn't it?"*
- #4095 (closed 2 days before this PR was opened): *"what if someone has another piped inlined that converts uuid to something else later?"*

The objection was to the idea of inferring OpenAPI schema from runtime pipes, not to implementation details, so I can't approve this on my own. The accepted direction in this repo is inference from TypeScript types / the CLI plugin (`Date` → `date-time`, `BigInt` → `int64`, node's `UUID` type in #2797, literal unions in #3964). Also note that #1818 never got a maintainer response, so it being open is not a signal that the feature is wanted.

This PR widens the scope of the declined design (3 more pipes) and, compared to #4095, handles the "multiple pipes" objection worse: it returns on the first matching pipe at any position in the chain, and the inferred type **overrides** the reflected TypeScript type. If you want to make the case for it, it needs a direct answer to that objection first. I'm leaving the final call to Kamil.

### Review of the implementation (in case the feature is reconsidered)

I ran the branch locally (unit tests: 520 passing, `tsc` clean). Issues found:

1. **Pipes are matched by `constructor.name`** (see inline). Verified locally: a user-land class that happens to be named `ParseIntPipe` flips the param from `string` to `number`, while `class MyIntPipe extends ParseIntPipe {}` is not detected at all. Minified builds silently lose the feature too.
2. **Contradictory schemas when merged with explicit decorators.** `@ApiParam({ name: 'id', type: Number })` + `@Param('id', ParseUUIDPipe)` now emits `{ "type": "number", "format": "uuid" }` (verified locally). The description says explicit decorators take precedence, but only key-by-key, and there is no test for that claim.
3. **It is not a no-op for existing apps**, even though it's flagged as non-breaking. Diffing the generated parameters against `master` for the same controller: every `ParseUUIDPipe` param gains `format: uuid` (code generators map that to a different type, e.g. `UUID`/`Guid`), params whose TS type disagrees with the pipe change type (`@Query('q', ParseFloatPipe) q: string` goes `string` → `number`), and params whose reflected type is `Object` (e.g. `id: string | number`) that used to be omitted now show up. Most of these are arguably *better* docs, but it's a document change people will see on a minor upgrade, so it would need to be called out (or be opt-in).
4. **`npx prettier --check` fails** on `lib/services/parameter-metadata-accessor.ts` and `test/services/parameter-metadata-accessor.spec.ts` (long signature line, trailing blank line at EOF).
5. **Test coverage gaps**: no test for `ParseFloatPipe`, for pipe *instances* (`new ParseIntPipe()` / `new ParseUUIDPipe({ version: '4' })`), for several pipes on one param (`new DefaultValuePipe(1), ParseIntPipe`), for a custom pipe before/after a built-in one, or for the precedence of `@ApiParam()`/`@ApiQuery()`.
6. Docs checkbox is unchecked: this would need a docs.nestjs.com PR (OpenAPI > Operations / Types and parameters).

### Related gaps (follow-up material, only if the feature is accepted)

- `ParseIntPipe` → `integer` would be more accurate than `number`.
- `new ParseIntPipe({ optional: true })` and `DefaultValuePipe` still emit `required: true` (and no `default`).
- `ParseEnumPipe` (→ `enum`), `ParseArrayPipe` (`new ParseArrayPipe({ items: Number })` still emits `items: { type: 'string' }`) and `ParseDatePipe` (→ `date-time`) are not covered, which makes the inference feel arbitrary from a user's perspective.

Requesting changes mainly because of the prior decisions above, not because of the nits.

Asadshah7950:
Thank you so much @micalevisk for the incredibly thorough review and for surfacing the historical context on #3829, #3864, and #4095.

I completely understand and agree with the architectural philosophy — runtime pipes operate in a sequential transformation pipeline where downstream pipes can alter wire types, making runtime pipe inspection inherently implicit compared to compile-time TypeScript / CLI plugin AST inference.

Given that Kamil and the team have consistently preferred keeping runtime pipe inspection manual to avoid these edge cases, I will close this PR. Really appreciate your detailed feedback and time reviewing!

